PR #5, opened 2020-07-11
Kotlin 1.4 M3

SalomonBrys:
- Updated Kotlin to 1.4-M3.
- Updated version number (0.2.0-1.4-M3).
- Moved BitcoinK into the `fr.acinq.bitcoink` package instead of `fr.acinq`, so that it conveys the same semantic & packaging as other Acinq libraries (also, in a file explorer, bitcoink artifacts are not mixed up with other acinq libs).
- Disabled granular source sets metadata as it makes dependency resolution fails in project depending on this since this lib does not use any granular source set.
- Updated secp256k1 version.
- Removed Hex object since it already exists in secp256k1 (removes code duplication).
- Updated CI to use the same mechanisms secp256k1-kmp uses.
- Updated Readme with installation instructions.


build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
 
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
-group = "fr.acinq"
+group = "fr.acinq.bitcoink"
```
sstone, line 15: What is the point of this change ? What kind of mixup does it prevent ?

SalomonBrys: In the bintray repository (as well as local of you `piblishToMavenLocal`), in the `fr/acinq` directory, you'll find the secp256k1 and eklair directories, which each contain their artifact directories (`-jvm`, `-iosX64`, `-metadata`, etc.).

If we leave bitcoink in the `fr.acinq` package, the `fr/acinq` directory will also contain bitcoink artifacts, and therefore expose a mixup of bitcoink artifacts and other library directories.

We should make a choice : either all our libraries are in the same `fr.acinq` package (which is the choice made, for example, by KotlinX), or each library is in its own package (which is also a widely used pattern).

Right now, we are using both patterns (the first in BitcoinK, the second in Secp251k1 and Eklair), which is confusing especially when browsing repository files. I think this is bad. We should have all our libraries embrace either one.

I personally prefer the second (each library in its own repository) as I find out easier to manage and understand, hence this PR.
However, as long as we are consistent with all our libraries, both choices are valid.

sstone: Ok I get it. Let's move bitcoink to its own package.

PR #17, opened 2021-03-23
Add method to compute addresses from public key scripts

sstone:


src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -38,27 +39,72 @@ public fun <T> List<T>.updated(i: Int, t: T): List<T> = when (i) {
...
+     */
+    @JvmStatic
+    public fun computeP2WpkhAddress(pub: PublicKey, chainHash: ByteVector32): String = pub.p2wpkhAddress(chainHash)
+
+    @JvmStatic
+    public fun computeBIP84Address(pub: PublicKey, chainHash: ByteVector32): String = computeP2WpkhAddress(pub, chainHash)
 
-public fun computeBIP84Address(pub: PublicKey, chainHash: ByteVector32): String = computeP2WpkhAddress(pub, chainHash)
+    @JvmStatic
+    public fun addressFromPublicKeyScript(chainHash: ByteVector32, pubkeyScript: ByteArray): String? {
```
t-bast, line 75: Would it make sense to also expose:

```
public fun addressFromPublicKeyScript(chainHash: ByteVector32, pubkeyScript: List<ScriptElt>): String?
```

instead of letting callers do `Script.write` themselves?

sstone: Yes it's a good idea, done in https://github.com/ACINQ/bitcoin-kmp/commit/bae9593b67474f98bc5e66ca346547c9a4324ed9

t-bast (approved):

PR #21, opened 2021-05-11
Add complete PSBT support

t-bast:
Add full support for PSBT v0 ([BIP 174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki)).
All official test vectors are passing and additional test vectors from [bitcoin-lib](https://github.com/acinq/bitcoin-lib) as well.
There is also a test showcasing how this can be used to bump the fees of a lightning anchor output from an external cold wallet.

The PSBT test vectors exposed a few bugs in other components that I fixed in this PR.


src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
```diff
@@ -0,0 +1,870 @@
...
+            val witnessScript: List<ScriptElt>?,
+            val derivationPaths: Map<PublicKey, KeyPathWithMaster>,
+            val unknown: List<DataEntry>
+        ) {
+            public companion object {
+                public val empty: PartiallySignedOutput = PartiallySignedOutput(null, null, mapOf(), listOf())
+            }
+        }
+
+        public sealed class UpdateResult {
```
sstone, line 365: Why introduce a new custom type instead of either:

- use exceptions (which the caller can wrap with `Result` or `Either`
- use `Either` explicitly

src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
```diff
@@ -0,0 +1,870 @@
...
+        }
+
+        private fun writeDataEntry(entry: DataEntry, output: Output) {
+            BtcSerializer.writeVarint(entry.key.size(), output)
+            output.write(entry.key.bytes)
+            BtcSerializer.writeVarint(entry.value.size(), output)
+            output.write(entry.value.bytes)
+        }
+
+        public sealed class ParsePsbtResult {
```
sstone, line 555: Same question as above

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -66,7 +66,17 @@ public object Script {
     @JvmStatic
     public fun parse(hex: String): List<ScriptElt> = parse(Hex.decode(hex))
 
-    public tailrec fun write(script: List<ScriptElt>, out: Output): Unit {
+    @JvmStatic
+    public fun parseSafe(blob: ByteArray): List<ScriptElt>? = try {
```
sstone, line 70: Why not let the caller use `try` or `runCatching` ?

src/commonMain/kotlin/fr/acinq/bitcoin/io/Input.kt
```diff
@@ -7,3 +7,4 @@ public interface Input {
 }
 
 public fun Input.readNBytes(n: Int): ByteArray = ByteArray(n).also { read(it, 0, n) }
+public fun Input.readNBytesStrict(n: Int): ByteArray? = if (availableBytes < n) null else readNBytes(n)
```
sstone, line 10: It's probably a bit off topic but we should probably just dump the "non strict" version of `readNBytes`

src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
t-bast: I don't want to rely on exceptions here (there's no good reason to use them) and we don't have `Either` available in bitcoin-kmp...that's why I went with the recommended way of writing Kotlin exception-free code.

But maybe it makes sense to import `Either` in `bitcoin-kmp` and use it here?

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: Because I want to avoid duplicate code since I'm using this in multiple places. But if you don't like this `parseSafe` I can revert that and be explicit where I catch exceptions.

src/commonMain/kotlin/fr/acinq/bitcoin/io/Input.kt
t-bast: I'll try that and will see; if that doesn't break anything in lightning-kmp it's worth doing!

src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
sstone: Yes I think that importing `Either` here is a good idea (I had forgotten that `Result` cannot be used as a return type...)

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/21/commits/41681ac80e554cdb895b4db5426aa160d3f3ef5d

src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/21/commits/f03bfb328fe8159c3ec5ed6014540e13643d025e

src/commonMain/kotlin/fr/acinq/bitcoin/Psbt.kt
t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/21/commits/f03bfb328fe8159c3ec5ed6014540e13643d025e

src/commonMain/kotlin/fr/acinq/bitcoin/io/Input.kt
t-bast: I'll do that in a separate PR, ok?

src/commonMain/kotlin/fr/acinq/bitcoin/utils/Either.kt
```diff
@@ -0,0 +1,65 @@
...
+    public data class Right<Nothing, out R>(val value: R) : Either<Nothing, R>() {
+        override val isLeft: Boolean = false
+        override val isRight: Boolean = true
+        override val left: Nothing? = null
+        override val right: R? = value
+    }
+}
+
+public inline fun <L, R, X> Either<L, R>.flatMap(f: (R) -> Either<L, X>): Either<L, X> = when (this) {
+    is Either.Left -> Either.Left(this.value)
```
sstone, line 53: Why not return `this` as we did in `lightning-kmp`? (it requires a cast but it's arguably a bit cleaner to not make a copy here)

sstone (commented):
LGTM! just one nit

src/commonMain/kotlin/fr/acinq/bitcoin/utils/Either.kt
t-bast: Exactly for that reason, I didn't like the unchecked cast...I don't think the potential extra copy (if not optimized away by the compiler) will hurt us to be honest.

sstone: It's not really the performance issue but the fact that it should be the same object (that also what Scala's `Either` does, with a cast :) see https://github.com/scala/scala/blob/2.13.x/src/library/scala/util/Either.scala#L353)

t-bast: That's a good point, it makes sense to keep the same object! I'll change that.

t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/21/commits/a2216b5fda9d9ba16eea305eeffd27577c77a55a

sstone (approved):

PR #25, opened 2021-06-25
Add support for Bech32m

t-bast:
This support will be necessary for future P2TR addresses.
We add forward-compatible support for segwit v1-16 as suggested in [BIP 350](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki).

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Bech32.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Bech32.kt
@@ -31,9 +31,9 @@ public typealias Int5 = Byte
 public object Bech32 {
     public const val alphabet: String = "qpzry9x8gf2tvdw0s3jn54khce6mua7l"
 
-    public sealed class Encoding {
-        public object Bech32 : Encoding()
-        public object Bech32m : Encoding()
+    public enum class Encoding(public val constant: Int) {
+        Bech32(1),
+        Bech32m(0x2bc830a3)
     }
 
     // char -> 5 bits value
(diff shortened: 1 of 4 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/Bech32.kt
```diff
@@ -19,16 +19,22 @@ package fr.acinq.bitcoin
...
+ * Bech32 and Bech32m address formats.
+ * See https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki and https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki.
+ */
 public object Bech32 {
     public const val alphabet: String = "qpzry9x8gf2tvdw0s3jn54khce6mua7l"
 
-    // 5 bits integer
-    // Bech32 works with 5bits values, we use this type to make it explicit: whenever you see Int5 it means 5bits values, and
-    // whenever you see Byte it means 8bits values
+    public sealed class Encoding {
```
sstone, line 34: Nit: why not use an enum instead, for example:
```scala
    public enum class Encoding(public val constant: Int) {
        Bech32(1),
        Bech32m(0x2bc830a3)
    }
```

t-bast: Good idea, I didn't know Kotlin had that option!

t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/25/commits/5d6d88b14eb17f915b14e9be8ae300a9461919a9

sstone (approved):

PR #26, opened 2021-06-29
Fix bug when reading "empty" transactions

sstone:
We were trying to read 4 bytes when only 2 were available, but this bug was hidden until we made our low-level primitives more strict.


src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -355,26 +355,30 @@ public data class Transaction(
...
+                /* The witness flag is present, and we support witnesses. */
+                flags = flags xor 1
+                val witnesses = mutableListOf<ScriptWitness>()
+                for (i in 0..tx.txIn.lastIndex) witnesses += ScriptWitness.read(input, protocolVersion)
+                tx = tx.updateWitnesses(witnesses.toList())
+                require(tx.hasWitness) { "Superfluous witness record" }
             }
+            require(flags == 0) { "Unknown transaction optional data" }
+            tx = tx.copy(lockTime = uint32(input).toLong())
+            return tx
```
t-bast, line 381: I'm not a fan of relying on mutating data here to avoid an `else` below, it makes the flow really hard to read.
Can't we do something like:

```suggestion
            return if ((flags and 1) != 0 && serializeTxWitness(protocolVersion)) {
                /* The witness flag is present, and we support witnesses. */
                flags = flags xor 1
                val witnesses = mutableListOf<ScriptWitness>()
                for (i in 0..tx.txIn.lastIndex) witnesses += ScriptWitness.read(input, protocolVersion)
                tx = tx.updateWitnesses(witnesses.toList())
                require(tx.hasWitness) { "Superfluous witness record" }
                tx.copy(lockTime = uint32(input).toLong())
            } else {
                require(flags == 0) { "Unknown transaction optional data" }
                tx.copy(lockTime = uint32(input).toLong())    
            }
```

t-bast (commented):
I added that test case to `bitcoin-lib` and it looks like we don't have the issue there, but I can't understand why, the code is pretty similar...

Maybe it's worth updating bitcoin-lib as well since you add more requirements to the flags that we previously had?

src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
sstone: It's true in general but here I made an exception because it makes the code almost identical to bitcoin core's which in some cases is a very good thing, as opposed to identical to bitcoin-lib's (which I believe has the same bug, hidden because java streams are very lenient).

t-bast (approved):

PR #45, opened 2022-01-19
Fix typo referencing wrong maven group id

0x8664b2:
Took me a second to figure out why the dependency couldn't be found.


t-bast (approved):
Good catch, thanks!

PR #46, opened 2022-02-07
Add segwit witness version helper

t-bast:
And a few missing JvmStatic annotations.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
@@ -40,6 +40,7 @@ public object Script {
      * @param stack initial command stack
      * @return an updated command stack
      */
+    @JvmStatic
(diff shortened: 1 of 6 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -266,6 +268,22 @@ public object Script {
     @JvmStatic
     public fun isNativeWitnessScript(script: ByteVector): Boolean = runCatching { parse(script) }.map { isNativeWitnessScript(it) }.getOrDefault(false)
 
+    private fun pushSize(op: ScriptElt): Int? = when (op) {
```
sstone, line 271: This could be useful, what about making it public ?

t-bast: Sure, why not, done in https://github.com/ACINQ/bitcoin-kmp/pull/46/commits/053f58ce1d6334321f84cb3524cb50844de7e436

sstone (approved):

PR #55, opened 2022-07-11
Fix signing psbt that spend p2wpkh outputs

sstone:
When signing an input that spends a P2WPKH output, the actual witness script that is used to compute the signature is the p2PKH script (for the same pubkey hash).


t-bast:
That PR is a bit misleading...it's not really fixing anything, it's just changing the flow of how p2wpkh inputs must be handled by callers. What you're really adding with this PR is the ability to handle p2wpkh inputs without having to explicitly provide the `witnessScript` when calling `updateWitnessInputTx`. Previously, we required the caller to explicitly provide `p2pkh(<pubkey>)` in the `witnessScript` field, which ensures they know what they're doing (that this input is a p2wpkh for the given <pubkey>). This was already extensively covered in tests (e.g. in `create psbt with various input types` and `manual coinjoin workflow`).

I'm a bit wary of doing this implicitly for them, I think it makes the caller code more obscure. An alternative would be to provide an explicit method `updatePay2wpkhInput()` instead of the generic `updateWitnessInputTx`, which would be clearer?

sstone:
I thought about this for a while and changed my mind: for now it is better to have users provide the correct witness script explicitly, they have to know how to derive the correct private key anyway and will know exactly what type of input they're signing.

PR #56, opened 2022-07-21
Add method to verify txout proofs

sstone:


src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -238,6 +241,67 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
...
+         * Verify a tx inclusion proof (a proof that a tx is included in a given block)
+         * @param proof tx inclusion proof, in the format used by bitcoin core's 'verifytxoutproof' RPC call
+         * @return a (block header, matched tx ids, matched tx positions) tuple
+         */
+        @JvmStatic
+        public fun verifyTxOutProof(proof: ByteArray): Triple<BlockHeader, List<ByteVector32>, List<Int>> {
+            val header = BlockHeader.read(proof.take(80).toByteArray())
+            val inputStream = ByteArrayInput(proof.drop(80).toByteArray())
+            val nTransactions = uint32(inputStream).toInt()
+            val vHash = readCollection(inputStream, { i, _ -> hash(i).byteVector32() }, null, Protocol.PROTOCOL_VERSION)
```
t-bast, line 254: What does the `v` prefix stand for?

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -238,6 +241,67 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
...
+            val vBits = script(inputStream)
+
+            fun bit(bits: ByteArray, pos: Int): Boolean {
+                val elt = bits[pos / 8]
+                return (elt.and((1.shl(pos % 8)).toByte()) != 0.toByte())
+            }
+
+            fun calcTreeWidth(nTransactions: Int, height: Int): Int = (nTransactions + (1.shl(height)) - 1).shr(height)
```
t-bast, line 262: This is very obscure...we should add comments to explain:

- the format of a proof
- what we're actually doing to verify it (in plain english)

Then it will be easier to review the code and verify that it does what we expect, because without that information this is very hard to read!

sstone: I've added comments and a link to the Bloom Filters BIPs (https://en.bitcoin.it/wiki/BIP_0037#Partial_Merkle_branch_format) which describes the format of partial  merkle trees.

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
sstone: I've removed it and switched to more conventional names in 1705727f6772bdcf6b9d9b00a42858fbf8505a43

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -238,6 +241,67 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
         @JvmStatic
         public fun checkProofOfWork(block: Block): Boolean = BlockHeader.checkProofOfWork(block.header)
 
+        /**
+         * Verify a tx inclusion proof (a proof that a tx is included in a given block)
+         * @param proof tx inclusion proof, in the format used by bitcoin core's 'verifytxoutproof' RPC call
+         * @return a (block header, matched tx ids, matched tx positions) tuple
+         */
+        @JvmStatic
+        public fun verifyTxOutProof(proof: ByteArray): Triple<BlockHeader, List<ByteVector32>, List<Int>> {
```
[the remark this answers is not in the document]
sstone: Yes it's better like this, done in 1705727f6772bdcf6b9d9b00a42858fbf8505a43

t-bast:
It took me a while to fully understand this (I wasted a lot of time because of the endianness of the bits encoding...), but now it all makes sense and the code is quite nice. I learnt how this worked by manually construction sample proofs in test cases, I think they would be useful to add to help others understand this mechanism so I created a commit for that: https://github.com/ACINQ/bitcoin-kmp/commit/4be1f2a2c86ff4afa3ebd1d53f4d100330a76fb6 (don't hesitate to change a few things after cherry-picking that commit if there are things you don't like).

I also added a few validation checks that were missing and slightly updated some comments for clarity (mostly, we must not forget to also check the header's proof of work).

sstone:
Thanks! I've added your changes with the exception of the checks on the number of hashes/bits used during verification. bitcoin core does not have these checks, and even if there are unused hashes/bits we still get valid proofs for the txids that we've matched.

t-bast (approved):

Issue #60, opened 2022-08-16
Add signet genesis block to Block.kt


remyers:
To cleanly add signet support to Eclair to address  [topic# 2382](https://github.com/ACINQ/eclair/discussions/2382) we need to add the signet genesis block to Block.kt in bitcoin-kmp.

The genesis block for Signet is described in the wiki entry for [BIP 0325](https://en.bitcoin.it/wiki/BIP_0325#Genesis_Block).

The hash should be `00000008819873e925422c1ff0f99f7cc9bbb232af63a077a480a3633bee1ef6` (see similar definition in c-lightning [here](https://github.com/rustyrussell/lightning/blob/5ac88cef90ef210fc9e12b3696a9f9507fe83fe6/bitcoin/chainparams.c#L85)).

sstone:
closed by https://github.com/ACINQ/bitcoin-kmp/pull/61

PR #61, opened 2022-08-16
Add support for signet genesis block

remyers:
See issue #60 and Eclair discussion/request [#2382](https://github.com/ACINQ/eclair/discussions/2382)

This PR also adds check of the final genesis block hashes.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
@@ -365,15 +365,6 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
         )
 
         @JvmField
-        public val SegnetGenesisBlock: Block = LivenetGenesisBlock.copy(
-            header = LivenetGenesisBlock.header.copy(
-                bits = 503447551,
-                time = 1452831101,
-                nonce = 0
-            )
-        )
-
-        @JvmField
         public val SignetGenesisBlock: Block = LivenetGenesisBlock.copy(
             header = LivenetGenesisBlock.header.copy(
                 bits = 503543726,
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
@@ -373,5 +373,13 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
             )
         )
 
+        @JvmField
+        public val SignetGenesisBlock: Block = LivenetGenesisBlock.copy(
+            header = LivenetGenesisBlock.header.copy(
+                bits = 503543726,
+                time = 1598918400,
+                nonce = 52613770
+            )
+        )
     }
 }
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -373,5 +373,13 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
             )
         )
 
+        @JvmField
+        public val SignetGenesisBlock: Block = LivenetGenesisBlock.copy(
```
t-bast, line 377: Do you know what the `Segnet` one above refers to?

t-bast (commented):
LGTM, just wondering what the `Segnet` one we have is supposed to be...

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
sstone: `Segnet` was a test network that was used to test segwit in 2016. It's most likely dead by now.

t-bast: Interesting, I guess we can remove it in this PR then? It's a bit confusing with `Signet`...

sstone: Agreed, it was ported from `bitcoin-lib` but is useless now.

t-bast:
@remyers can you remove the `Segnet` hash in this PR as well (and in https://github.com/ACINQ/bitcoin-lib/pull/70)?

remyers:
> @remyers can you remove the `Segnet` hash in this PR as well (and in [ACINQ/bitcoin-lib#70](https://github.com/ACINQ/bitcoin-lib/pull/70))?

It looks like I will have to recreate on `testnet` some tests in [SegwitTestsCommon.kt](https://github.com/ACINQ/bitcoin-kmp/blob/e69e17c236359d6e164eb5c9ea3837f632be86af/src/commonTest/kotlin/fr/acinq/bitcoin/SegwitTestsCommon.kt#L117) that reference `segnet` transactions.

t-bast:
> It looks like I will have to recreate on testnet some tests in [SegwitTestsCommon.kt](https://github.com/ACINQ/bitcoin-kmp/blob/e69e17c236359d6e164eb5c9ea3837f632be86af/src/commonTest/kotlin/fr/acinq/bitcoin/SegwitTestsCommon.kt#L117) that reference segnet transactions.

Good point, but I think most of these tests are obsolete, and were only useful before segwit was activated. You can probably greatly simplify this test file by just verifying that p2wpkh and p2wsh work as expected, without referencing segnet transactions, this will remove some technical debt!

remyers:
The tests seemed useful, so I just ported them to testnet in [2581f47](https://github.com/ACINQ/bitcoin-kmp/pull/61/commits/2581f47815f647aa1b7fa1f8ad8e27719efba2f1) after removing segnet.

t-bast (approved):
LGTM :100: 

segwit

Yay saved fees!

PR #64, opened 2022-09-21
Use kotlin 1.6

sstone:


build.gradle.kts
```diff
@@ -49,7 +49,7 @@ kotlin {
                 implementation(kotlin("test-common"))
                 implementation(kotlin("test-annotations-common"))
                 implementation("org.kodein.memory:kodein-memory-files:0.7.0")
```
dpad85, line 51: `kodein-memory:0.7.0` is very old (Feb 2021), it would be a good opportunity to upgrade to version 0.11.0 (0.12.0 supports kotlin 1.7).

sstone: Since it's used for testing we could upgrade later (when we move to kotlin 1.7 for example) ?

dpad85 (approved):
👍

PR #69, opened 2022-10-04
Improve Tapscript Core tests

t-bast:
Most tests (~90%) contain both a success and a failure test case. We previously only tested the success case when that happened.


sstone (approved):
Nice!

PR #70, opened 2022-10-06
Update Schnorr Signing API

sstone:


t-bast (approved):
LGTM

remyers:
What do you think of my [suggestion](https://github.com/ACINQ/bitcoin-kmp/pull/40#discussion_r989924833) from the parent PR?
 
Basically I suggest  `signSchnorr` takes an argument with the `SchnorrTweak` interface. We also define a `TapTweak` sealed class which implements the `SchnorrTweak` interface.   This way we can  pass a `TapTweak` to the `outputKey` function which does not include the possbility for `NoTweak`. We can also in the future define different tweaks with the `SchnorrTweak` interface used for Schnorr signing (eg. for PTLCs).

PR #71, opened 2022-10-10
Update schnorr signing api

sstone:


t-bast (approved):

src/commonMain/kotlin/fr/acinq/bitcoin/Crypto.kt
```diff
@@ -154,22 +154,40 @@ public object Crypto {
         )
     }
 
+    /**
+     * Specify how private keys are tweaked when creating Schnorr signatures
```
remyers, line 158: These tweaks can be applied to both private and public keys (via outputKey) so comments should probably just refer to "key" not "private key".

remyers (approved):
Looks good! this one is still my preferred option to be maximally pendantic.

PR #73, opened 2022-10-18
Make script execution faster

sstone:
It was very inefficient on native platforms, with native tests taking up to 1h on ubuntu and osx64 instead of 30 seconds on the JVM. The root cause was that native kotlin collections are very inefficient and some of the patterns we imported from Scala do not work. For example calling tail() on immutable lists is very inefficient (which is suprising coming from Scala). Here we just use mutable collections in the script execution loop, which is not recursive anymore.


t-bast (commented):
The code is much clearer with those changes than with the functional callbacks!
It looks good to me, there's no functional change with what we had before and it was a good opportunity to review our stack manipulation code.

I have some tweaks proposed in #74

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -724,629 +716,488 @@ public object Script {
...
+                        conditions[0] = !conditions[0]
+                        opCount++
+                    }
+                    head == OP_ENDIF -> {
+                        conditions.removeAt(0)
+                        opCount++
+                    }
+                    conditions.any { !it } -> {
+                        opCount += cost(head)
+                    }
```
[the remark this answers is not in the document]
sstone: Ah yes this one is tricky: `conditions` is a stack of all booleans that were checked by previous OP_IF and OP_NOTIF instructions. 
`conditions.any { !it }` means that we're in a IF/NOTIF branch that is not executed, so we don't even check what the current instruction is, but it still counts as one instruction. I'll add a comment with an example.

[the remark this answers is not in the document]
sstone: see https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/9f4b81a0ae141a0ed2711c9bf05e3fa1d32c7ae3. it was also an opportunity to revisit and simplify how we count opcode and check opcode limits.

src/commonTest/kotlin/fr/acinq/bitcoin/TaprootTestsCommon.kt
```diff
@@ -348,4 +351,17 @@ class TaprootTestsCommon {
...
     }
+
+    @Test
+    fun `parse and validate huge transaction`() {
+        // this is the tx that broke btcd/lnd
+        val file = resourcesDir().resolve("7393096d97bfee8660f4100ffd61874d62f9a65de9fb6acf740c4c386990ef73.bin").openReadableFile()
+        val buffer = ByteArray(file.available)
+        file.readBytes(buffer)
+        file.close()
+        val tx = Transaction.read(buffer)
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/43e80b21f78111526bd2c39828d905c734e11163

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -930,15 +934,14 @@ public object Script {
...
-                        opCount++
                     }
 
                     head == OP_CHECKMULTISIG || head == OP_CHECKMULTISIGVERIFY -> {
                         require(signatureVersion != SigVersion.SIGVERSION_TAPSCRIPT) { "invalid OP_CHECKMULTISIG operation" }
                         // pop public keys
                         val m = decodeNumber(stack.removeFirst()).toInt()
                         if (m < 0 || m > 20) throw RuntimeException("OP_CHECKMULTISIG: invalid number of public keys")
-                        val nextOpCount = opCount + 1 + m
+                        val nextOpCount = opCount  + m
```
t-bast, line 944: We don't really need to introduce `nextOpCount`, we could simply do `opCount += m` here?

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -1267,7 +1232,6 @@ public object Script {
                         stack[1] = x3
                         stack[2] = x0
                         stack[3] = x1
-                        opCount++
                     }
 
                     head == OP_TOALTSTACK -> {
```
t-bast, line 1237: Note that this one previously didn't increment the `opCount`, and now it does. Is that expected? Did we have a bug here?

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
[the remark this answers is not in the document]
t-bast: Thanks, the explanations for the `conditions` list are very clear, it makes a lot of sense now!

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
sstone: Yes it's better like this, done in https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/491f099bf1fdc0fb1e2b30721396eed1ef3fb175

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
sstone: I saw this one too, and it does look like a bug :( (not 100% sure yet).

t-bast (approved):
ACK https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/408cf3dcf70fa743a4e1f0404942f31e764bf7cb, this is much better!

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: It shouldn't be too much of an issue, as in this case we are more lenient than bitcoind: we will accept some transactions that will be rejected by bitcoind. This cannot be exploited in our lightning usage, since we wait for txs to confirm, unless we're the only tx funder (in which case we won't create such invalid txs).

remyers (approved):
I reviewed against `EvalScript` in core and did not see any changes from previous behavior. I focused more deeply on the conditional logic and also agree these changes fix an `opCount` issue. 

This was easier to review against core then the recursive version and seems like a win-win overall. 

Reviewing script consensus logic is hard though, so we should keep our eye on efforts in core to create a consensus library that we could use instead of our own implementation.

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -724,629 +716,488 @@ public object Script {
...
             signatureVersion: Int
         ): List<ByteVector> {
             stack.forEach { require(it.size() <= MAX_SCRIPT_ELEMENT_SIZE) { "item is bigger than maximum push size" } }
-            return runInternal(script.withIndex().toList(), stack, state, signatureVersion)
+            return runInternal(script, stack, signatureVersion)
         }
 
         /**
          * internal execution loop.
          * it uses an indexed list of script elements because we may need to know the exact position of an element in the original script
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/10d7372be3febb443914105a97dd4f61dadeb331

src/commonTest/kotlin/fr/acinq/bitcoin/reference/ScriptTestsCommon.kt
```diff
@@ -55,6 +55,26 @@ class ScriptTestsCommon {
...
+           ]
```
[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/73/commits/10d7372be3febb443914105a97dd4f61dadeb331

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -785,10 +785,11 @@ public object Script {
                         }
                     }
 
-                    head == OP_NOTIF && conditions.any { !it } -> {
+                    op == OP_NOTIF && conditions.any { !it } -> {
                         conditions.add(0, true)
                     }
-                    head == OP_NOTIF -> {
+                    op == OP_NOTIF && stack.isEmpty() -> throw RuntimeException("Invalid OP_IF construction")
```
t-bast, line 791: nit:
```suggestion
                    op == OP_NOTIF && stack.isEmpty() -> throw RuntimeException("Invalid OP_NOTIF construction")
```

Also `OP_ELSE` and `OP_ENDIF` have the same copy-paste issue.

sstone: Bitcoin core uses the same error message for IF/NOTIF/... errors but you're right it's better to be more specific

t-bast (approved):
LGTM :rocket:

remyers (approved):
Looks great!

PR #82, opened 2023-01-30
Add method to modify PSBT unknown fields

sstone:


sstone:
Not needed after all.

PR #84, opened 2023-02-22
Add weight() to TxIn and TxOut

t-bast:
This is useful when sharing on-chain fees in the interactive-tx protocol used in lightning for dual funding and splicing.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
@@ -209,6 +209,7 @@ public data class TxIn(
 
         @JvmStatic
         public fun weight(txIn: TxIn, protocolVersion: Long = PROTOCOL_VERSION): Int {
+            // Note that the write function doesn't serialize witness data, so we count it separately.
             val witnessWeight = if (txIn.hasWitness) ScriptWitness.write(txIn.witness, protocolVersion).size else 0
             return 4 * write(txIn).size + witnessWeight
         }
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -204,6 +206,12 @@ public data class TxIn(
 
         @JvmStatic
         public fun coinbase(script: List<ScriptElt>): TxIn = coinbase(Script.write(script))
+
+        @JvmStatic
+        public fun weight(txIn: TxIn, protocolVersion: Long = PROTOCOL_VERSION): Int {
+            val witnessWeight = if (txIn.hasWitness) ScriptWitness.write(txIn.witness, protocolVersion).size else 0
+            return 4 * write(txIn).size + witnessWeight
```
sstone, line 213: nit: this is correct but difficult to compare to how we compute weight for transactions (`3 * baseSize + totalSize`) unless you remember that `write(txIn)` does not include `txIn`'s witness data, maybe a comment would help ?

t-bast: Definitely! I had a hard time understanding why we did `3 * baseSize + totalSize` for the whole transaction, I'll add a comment there as well.

t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/84/commits/470a1180c576295bd05858e07d780ab549d58d84

src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -430,16 +447,21 @@ public data class Transaction(
...
+        /** Total size of the transaction with witness data, if any. */
         @JvmStatic
         public fun totalSize(tx: Transaction, protocolVersion: Long = PROTOCOL_VERSION): Int = write(tx, protocolVersion).size
 
         @JvmStatic
-        public fun weight(tx: Transaction, protocolVersion: Long = PROTOCOL_VERSION): Int =
-            totalSize(tx, protocolVersion) + 3 * baseSize(tx, protocolVersion)
+        public fun weight(tx: Transaction, protocolVersion: Long = PROTOCOL_VERSION): Int {
+            // Witness data uses 1 weight unit, while non-witness data uses 4 weight units.
+            // We thus serialize ones with witness data and 3 times without witness data.
```
sstone, line 462:
```suggestion
            // We thus serialize once with witness data and 3 times without witness data.
```

sstone (approved):

PR #86, opened 2023-03-08
Allow a segwit PSBT input to be updated with a non-segwit utxo

sstone:
Non-segwit utxos incude the full tx being spent and not just its output.
This can be used to detect "fee attacks" where the segwit utxo's output amount does  not match the amount that is actually spent.


src/commonTest/kotlin/fr/acinq/bitcoin/psbt/PsbtTestsCommon.kt
```diff
@@ -1080,6 +1084,72 @@ class PsbtTestsCommon {
...
+        val psbt3 = Psbt(tx)
+            .updateWitnessInput(OutPoint(utxo1, 0), utxo1.txOut[0], null, Script.pay2pkh(priv1.publicKey()))
+            .flatMap { it.updateNonWitnessInput(utxo1, 0) }
+            .flatMap { it.updateWitnessInput(OutPoint(utxo2, 0), utxo2.txOut[0].updateAmount(Satoshi(5_000_000L + 1000)), null, Script.pay2pkh(priv2.publicKey())) }
+            .flatMap { it.updateNonWitnessInput(utxo2, 0) }
+            .right!!
+        assertEquals(psbt3.inputs[1].witnessUtxo!!.publicKeyScript, psbt3.inputs[1].nonWitnessUtxo!!.txOut[0].publicKeyScript)
+        // we can see that the witness utxo amount does not match the amount of the corresponding output in the non-witness utxo
+        assertNotEquals(psbt3.inputs[1].witnessUtxo!!.amount, psbt3.inputs[1].nonWitnessUtxo!!.txOut[0].amount)
+    }
```
remyers, line 1151: I tried to test that an invalid psbt would throw the appropriate error when deserialized, but it instead fails to deserialize the nonWitnessUtxo, any ideas why?
 
```
        // a psbt with mismatched nonWitnessUtxo and witnessUtxo should not deserialize
        assertEquals(Psbt.read(Psbt.write(psbt3)), Either.Left(ParseFailure.InvalidTxInput("non-witness utxo does not match psbt outpoint")))
```

remyers (changes requested):
Changes look good, but I found an issue when I tried to serialize/deserialize the psbt with invalid nonWitness input data.

You should also add tests for the `derivationPaths` being properly updated.

src/commonTest/kotlin/fr/acinq/bitcoin/psbt/PsbtTestsCommon.kt
sstone: it seems that Psbt.write(psbt3) produces an invalid ouput, I don't know why yet

sstone: fixed in https://github.com/ACINQ/bitcoin-kmp/pull/86/commits/661a6765e140cf014dfb6065390bd064e5c4b056
(the non-witness utxo was a tx without inputs that could not be de-serialised, I've added dummy input)

remyers (approved):
LGTM! these changes will help inform review of [PR #2345](https://github.com/ACINQ/eclair/pull/2345/files)

src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
```diff
@@ -198,7 +198,12 @@ public data class Psbt(val global: Global, val inputs: List<Input>, val outputs:
                 sighashType = sighashType ?: input.sighashType,
                 derivationPaths = input.derivationPaths + derivationPaths
             )
-            is Input.WitnessInput.PartiallySignedWitnessInput -> return Either.Left(UpdateFailure.CannotUpdateInput(inputIndex, "cannot update non-segwit input: it has already been updated with segwit data"))
+            is Input.WitnessInput.PartiallySignedWitnessInput -> input.copy(
+                nonWitnessUtxo = inputTx,
+                redeemScript = redeemScript ?: input.redeemScript,
+                sighashType = sighashType ?: input.sighashType,
+                derivationPaths = input.derivationPaths + derivationPaths
+            )
```
t-bast, line 206: This is very confusing now, we allow mixing segwit and non-segwit functions, which is a terrible API for callers...if we decide we only want to support scenarios where the full previous transaction is always provided, we should make deeper changes to make it much clearer for callers (and that would greatly simplify the code as well).

sstone: Yes naming is a bit confusing but the change here is very simple and matches what bitcoin core does: they  add a `non_witness_utxo` field to all PSBT segwit v0 inputs for the same reason: to protect against fee attacks (see https://github.com/bitcoin/bitcoin/blob/v24.0.1/src/wallet/wallet.cpp#L2059).

PR #90, opened 2023-04-04
Introduce `TxId` and `TxHash` types

t-bast:
It was really footgunny to only use a `ByteVector32` here, as users could use a `TxHash` when a `TxId` was expected. Introducing explicit types ensures that functions can clearly specify which one they expect.

Codecs can then decide how to encode that on the wire: this will let us exclusively use `TxId` in lightning, and encode/decode it as a `TxHash` on the wire, which will greatly simplify our types. For example, we would define `splice_locked` as:

```scala
data class SpliceLocked(val channelId: ByteVector32, val fundingTxId: TxId)
```

And the codec would convert this to a `TxHash` before encoding its bytes, since lightning wire messages always expect a `TxHash`. This way we hide that encoding detail in the codecs, and only deal with `TxId`s that we get from `Transaction` objects.


src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -25,32 +25,54 @@ import kotlin.jvm.JvmField
...
-    /**
-     * @return the id of the transaction this output belongs to
-     */
     @JvmField
-    public val txid: ByteVector32 = hash.reversed()
+    public val txid: TxId = TxId(hash)
 
     public val isCoinbase: Boolean get() = isCoinbase(this)
 
+    override fun toString(): String = "$txid:$index"
```
t-bast, line 71: I hesitated between using the `txid` or the `hash` here, what do you think? We're using the `txid` in most logs and that's what explorers and public-facing RPCs expect, so I figured it was better that way, but I could be wrong.

build.gradle.kts
```diff
@@ -12,7 +12,7 @@ plugins {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 group = "fr.acinq.bitcoin"
-version = "0.11.1"
+version = "0.12.0-SNAPSHOT"
```
t-bast, line 15: Is that what I'm supposed to do @sstone?

sstone: Yes

sstone (commented):
It can help, though I think that the biggest issue is not the lack of specific types but improper use of `tx hash/tx id` in many technical docs and specs (not ours :)). For consistency's sake we should also add new `block hash/block id` types

t-bast:
> For consistency's sake we should also add new block hash/block id types

Done in https://github.com/ACINQ/bitcoin-kmp/pull/90/commits/b9b8397748184a4e1c7fee7f750cc4cbe4963a09
I'm not completely sure what to use for `hashPreviousBlock`, there is no unit test to help me figure it out.
I guess it should use `BlockHash`, but how can I verify that?

sstone:
> > For consistency's sake we should also add new block hash/block id types
> 
> Done in [b9b8397](https://github.com/ACINQ/bitcoin-kmp/commit/b9b8397748184a4e1c7fee7f750cc4cbe4963a09) I'm not completely sure what to use for `hashPreviousBlock`, there is no unit test to help me figure it out. I guess it should use `BlockHash`, but how can I verify that?

We could add a test with 2 consecutive testnet blocks ?

t-bast:
> We could add a test with 2 consecutive testnet blocks ?

Done in https://github.com/ACINQ/bitcoin-kmp/pull/90/commits/7680d1f6626e288df0424251513c4c1e12ac2a26, looks like this was indeed using the block hash.

t-bast:
Rebased.

pm47 (approved):
Concept ACK.

t-bast:
Added to eclair in https://github.com/ACINQ/eclair/pull/2742 and lightning-kmp in https://github.com/ACINQ/lightning-kmp/pull/524 without any issue, it should be safe to move forward.

PR #94, opened 2023-07-03
Use proper types for address and script conversion errors

sstone:
We add specific return types to handle invalid address and public key script conversions. 
This is better than throwing exceptions because it makes easier for clients to create their own error handling code with localized error messages.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
@@ -16,7 +16,6 @@
 
 package fr.acinq.bitcoin
 
-import fr.acinq.bitcoin.utils.Either
 import kotlin.jvm.JvmStatic
 
 public const val MaxBlockSize: Int = 1000000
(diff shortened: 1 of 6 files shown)


dpad85 (approved):
LGTM

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -26,6 +27,20 @@ public fun <T> List<T>.updated(i: Int, t: T): List<T> = when (i) {
     else -> this.take(i) + t + this.drop(i + 1)
 }
 
+public sealed class InvalidAddress {
```
t-bast, line 30: If we want to avoid exposing `Either`, the easy solution would be to create an `AddressFromPublicKeyScriptResult` return type:

```
public sealed class AddressFromPublicKeyScriptResult {
  data class Success(val address: String) : AddressFromPublicKeyScriptResult()
  sealed class InvalidAddress : AddressFromPublicKeyScriptResult() {
    ...
  }
  sealed class InvalidPublicKeyScript : AddressFromPublicKeyScriptResult() {
    ...
  }
}

public fun addressFromPublicKeyScript(chainHash: ByteVector32, pubkeyScript: List<ScriptElt>): AddressFromPublicKeyScriptResult
```

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -26,6 +26,76 @@ public fun <T> List<T>.updated(i: Int, t: T): List<T> = when (i) {
     else -> this.take(i) + t + this.drop(i + 1)
 }
 
+public sealed interface AddressToPublicKeyScriptResult {
```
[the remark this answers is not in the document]
sstone: Extending `Throwable` makes implementing `getOrThrow()` or converting to `Either`-style types easier but it's debatable.
I think we need getter for scala interop (?) I'll check.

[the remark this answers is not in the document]
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/94/commits/41e2113a2dc556d8970a267c4c917bfd938804e8, I checked that it is usable from `bitcoin-lib`

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -56,71 +115,80 @@ public object Bitcoin {
     @JvmStatic
     public fun computeBIP84Address(pub: PublicKey, chainHash: ByteVector32): String = computeP2WpkhAddress(pub, chainHash)
 
+    /**
+     * Compute an address form a public key script
```
t-bast, line 119: nit:
```suggestion
     * Compute an address from a public key script
```

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -56,71 +115,80 @@ public object Bitcoin {
...
 
     @JvmStatic
-    public fun addressFromPublicKeyScript(chainHash: ByteVector32, pubkeyScript: ByteArray): String? {
+    public fun addressFromPublicKeyScript(chainHash: ByteVector32, pubkeyScript: ByteArray): AddressFromPublicKeyScriptResult {
         return try {
             addressFromPublicKeyScript(chainHash, Script.parse(pubkeyScript))
-        } catch (e: Exception) {
-            null
+        } catch (t: Throwable) {
+            AddressFromPublicKeyScriptResult.Failure.GenericError(t)
```
t-bast, line 186: If we have an exception here, it means that `Script.parse(pubkeyScript)` failed (since all other exceptions are already handled in the call to the overloaded `addressFromPublicKeyScript`), so we can probably use a dedicated error?

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -144,29 +212,37 @@ public object Bitcoin {
...
+                            it.first == "bc" && chainHash == Block.LivenetGenesisBlock.hash -> AddressToPublicKeyScriptResult.Success(listOf(witnessVersion, OP_PUSHDATA(it.third)))
+                            it.first == "tb" && chainHash == Block.TestnetGenesisBlock.hash -> AddressToPublicKeyScriptResult.Success(listOf(witnessVersion, OP_PUSHDATA(it.third)))
+                            it.first == "tb" && chainHash == Block.SignetGenesisBlock.hash -> AddressToPublicKeyScriptResult.Success(listOf(witnessVersion, OP_PUSHDATA(it.third)))
+                            it.first == "bcrt" && chainHash == Block.RegtestGenesisBlock.hash -> AddressToPublicKeyScriptResult.Success(listOf(witnessVersion, OP_PUSHDATA(it.third)))
+                            else -> AddressToPublicKeyScriptResult.Failure.ChainHashMismatch
                         }
                     },
                     onFailure = {
-                        error("$address is neither a valid Base58 address ($base58error) nor a valid Bech32 address ($it)")
+                        AddressToPublicKeyScriptResult.Failure.GenericFailure(it)
```
t-bast, line 245: We can probably use a more specific error than `GenericFailure` here, the address string is invalid or its encoding unknown, maybe an `InvalidAddress` error?

sstone: Yes, done in https://github.com/ACINQ/bitcoin-kmp/pull/94/commits/7f37fe904ef42b0fa1b8b45e0324e253674ec843

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/94/commits/7f37fe904ef42b0fa1b8b45e0324e253674ec843

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
sstone: Fixed in https://github.com/ACINQ/bitcoin-kmp/pull/94/commits/7f37fe904ef42b0fa1b8b45e0324e253674ec843

t-bast (approved):

PR #107, opened 2023-12-14
Add high-level types for musig2 primitives provided by secp256k1-kmp

sstone:
This PR is based on https://github.com/ACINQ/bitcoin-kmp/pull/108 because of https://github.com/ACINQ/secp256k1-kmp/pull/93


t-bast:
Why did you keep the prototype musig2 implementation (in the test folders)? Shouldn't we get rid of it entirely, we want to test the "real" implementation and have no use for the prototype one anymore?

sstone:
I kept it for reference but it's not supposed to be used in `Musig2TestsCommon.kt` I'll fix that. 
I don't think it's too soon to start reviewing this PR as the API in https://github.com/ACINQ/secp256k1-kmp/pull/93 should not change much (though we still have a few weeks before it gets merged).

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    public companion object {
+        /**
+         * @param pubkeys public keys to aggregate
+         * @param cache an optional key aggregation cache
+         * @return a new (if cache was null) or updated cache, and the aggregated public key
+         */
+        @JvmStatic
+        public fun add(pubkeys: List<PublicKey>, cache: KeyAggCache? = null): Either<Throwable, Pair<XonlyPublicKey, KeyAggCache>> = try {
```
t-bast, line 46: This doesn't need to return an `Either`: if all public keys are valid (which callers should have checked beforehand, because otherwise basically nothing works), I believe that key aggregation cannot fail.

In my branch I have added a comment that the caller needs to verify that public keys are valid, and a `require` at the beginning of the function:

```
require(publicKeys.all { it.isValid() }) { "some of the public keys provided are not valid" }
```

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    public fun toByteArray(): ByteArray = data.toByteArray()
+
+    /**
+     * @param secretNonce secret nonce
+     * @param pk private key
+     * @param aggCache key aggregation cache
+     * @return a Musig2 partial signature
+     */
+    public fun sign(secretNonce: SecretNonce, pk: PrivateKey, aggCache: KeyAggCache): Either<Throwable, ByteVector32> = try {
```
t-bast, line 72: This doesn't need to return an `Either`: the caller isn't malicious so its secret nonce and private key will be valid, and in that case I believe that the signing operation cannot fail.

If the caller is providing an invalid secret nonce or private key, it's ok to throw an exception because they're clearly doing something very wrong.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    public companion object {
+        /**
+         * @param aggregatedNonce aggregated public nonce
+         * @param msg message to sign
+         * @param cache key aggregation cache
+         * @return a Musig signing session
+         */
+        @JvmStatic
+        public fun build(aggregatedNonce: AggregatedNonce, msg: ByteVector32, cache: KeyAggCache): Either<Throwable, Session> = try {
```
t-bast, line 110: I don't think this needs to return an `Either`. It can only fail if the aggregated nonce is invalid, or the key aggregation cache is invalid. All of those will have been checked beforehand by the caller (because nonce aggregation and key aggregation return an `Either`), so we can assume that they're valid (otherwise the caller is clearly doing something very buggy and it's ok to throw).

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -0,0 +1,191 @@
...
+         * @param sessionId random session id. Must not be reused !!
+         * @param seckey optional private key
+         * @param pubkey public key
+         * @param msg optional message to sign
+         * @param cache optional key aggregation cache
+         * @param extraInput optional extra input value
+         * @return a (secret nonce, public nonce) tuple
+         */
+        @JvmStatic
+        public fun generate(sessionId: ByteVector32, seckey: PrivateKey?, pubkey: PublicKey, msg: ByteVector32?, cache: KeyAggCache?, extraInput: ByteVector32?): Either<Throwable, Pair<SecretNonce, IndividualNonce>> = try {
```
t-bast, line 142: This cannot fail either unless the caller is buggy and provides invalid arguments (e.g. a `pubkey` that doesn't match `seckey`). This thus doesn't need to return an `Either`.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: It will fail if the nonce, cache, or session is invalid, so I agree we don't need to return `Either` here

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/commit/bc6e398197b58cd73449d80e53e8f31f7d79375f

sstone: done in https://github.com/ACINQ/bitcoin-kmp/commit/bc6e398197b58cd73449d80e53e8f31f7d79375f

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/commit/bc6e398197b58cd73449d80e53e8f31f7d79375f

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/commit/bc6e398197b58cd73449d80e53e8f31f7d79375f

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/commit/bc6e398197b58cd73449d80e53e8f31f7d79375f

PR #109, opened 2024-01-12
Provide high-level taproot abstractions

t-bast:
We provide helpers and better high-level abstractions for spending taproot outputs via either the key path or script path. This ensures that callers can't mess up which `SigVersion` they use, or forget to include the script leaf hash or merkle root depending on what they are spending.

The caller only needs to remember:

- the internal public key
- the script tree
- if they're using the script path, the ID of the script leaf they're using

That is information that they should always be able to keep around or recompute, and seems to be minimal.

NB: the two commits are somewhat independent and should be reviewed separately.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
@@ -386,6 +386,20 @@ public object Script {
     @JvmStatic
     public fun isPay2wsh(script: ByteArray): Boolean = isPay2wsh(parse(script))
 
+    @JvmStatic
+    public fun isPay2tr(script: List<ScriptElt>): Boolean {
+        return when {
+            script.size == 2 && script[0] == OP_1 && script[1].isPush(32) -> true
+            else -> false
+        }
+    }
+
+    @JvmStatic
+    public fun isPay2tr(script: ByteArray): Boolean = isPay2tr(parse(script))
+
+    @JvmStatic
+    public fun isPay2tr(script: ByteVector): Boolean = isPay2tr(script.toByteArray())
+
     /**
      * @param pubKeyHash public key hash
      * @return a pay-to-public-key-hash script
@@ -404,7 +418,6 @@ public object Script {
     public fun pay2pkh(pubKey: PublicKey): List<ScriptElt> = pay2pkh(pubKey.hash160())
 
     /**
-     *
      * @param script bitcoin script
      * @return a pay-to-script script
      */
@@ -412,7 +425,6 @@ public object Script {
     public fun pay2sh(script: List<ScriptElt>): List<ScriptElt> = pay2sh(write(script))
 
     /**
-     *
      * @param script bitcoin script
      * @return a pay-to-script script
      */
@@ -420,7 +432,6 @@ public object Script {
     public fun pay2sh(script: ByteArray): List<ScriptElt> = listOf(OP_HASH160, OP_PUSHDATA(Crypto.hash160(script)), OP_EQUAL)
 
     /**
-     *
      * @param script bitcoin script
      * @return a pay-to-witness-script script
      */
@@ -428,7 +439,6 @@ public object Script {
     public fun pay2wsh(script: List<ScriptElt>): List<ScriptElt> = pay2wsh(write(script))
 
     /**
-     *
      * @param script bitcoin script
      * @return a pay-to-witness-script script
      */
@@ -436,7 +446,6 @@ public object Script {
     public fun pay2wsh(script: ByteArray): List<ScriptElt> = listOf(OP_0, OP_PUSHDATA(Crypto.sha256(script)))
 
     /**
-     *
      * @param script bitcoin script
      * @return a pay-to-witness-script script
      */
@@ -444,7 +453,6 @@ public object Script {
     public fun pay2wsh(script: ByteVector): List<ScriptElt> = pay2wsh(script.toByteArray())
 
     /**
-     *
      * @param pubKeyHash public key hash
      * @return a pay-to-witness-public-key-hash script
      */
@@ -455,19 +463,11 @@ public object Script {
     }
 
     /**
-     *
      * @param pubKey public key
      * @return a pay-to-witness-public-key-hash script
      */
     @JvmStatic
     public fun pay2wpkh(pubKey: PublicKey): List<ScriptElt> = pay2wpkh(pubKey.hash160())
-
-    /**
-     * @param pubkey x-only public key
-     * @return a pay-to-taproot script
-     */
-    @JvmStatic
-    public fun pay2tr(pubkey: XonlyPublicKey): List<ScriptElt> = listOf(OP_1, OP_PUSHDATA(pubkey.value))
 
     /**
      * @param pubKey public key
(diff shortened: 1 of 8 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -612,19 +612,21 @@ public object Script {
...
+         * @param internalPubKey internal public key.
+         * @param scriptTree tapscript tree.
+         * @param spendingScript script leaf we're spending (must exist in the [scriptTree]).
          */
         @JvmStatic
-        public fun build(internalPubKey: XonlyPublicKey, merkleRoot: ByteVector32, leaves: List<ScriptTree<ScriptLeaf>> = listOf()): ByteArray {
-            val parity = internalPubKey.outputKey(Crypto.TaprootTweak.ScriptTweak(merkleRoot)).second
-            val initialBlock = byteArrayOf((TAPROOT_LEAF_TAPSCRIPT + (if (parity) 1 else 0)).toByte()) + internalPubKey.value.toByteArray()
-            return leaves.fold(initialBlock) { block, tree -> block + ScriptTree.hash(tree).toByteArray() }
+        public fun build(internalPubKey: XonlyPublicKey, scriptTree: ScriptTree, spendingScript: ScriptTree.Leaf): ByteVector {
```
sstone, line 665: We should also allow users to build a control block even if they don't have the complete script tree but just its merkle root and a a merkle proof

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -477,6 +477,47 @@ public object Script {
...
+    @JvmStatic
+    public fun pay2tr(outputKey: XonlyPublicKey): List<ScriptElt> = listOf(OP_1, OP_PUSHDATA(outputKey.value))
+
+    /**
+     * @param internalKey internal public key that will be tweaked with the [scripts] provided.
+     * @param scripts optional spending scripts that can be used instead of key-path spending.
+     * @return the script and the tweak that must be applied to the private key for [internalKey] when signing.
+     */
+    @JvmStatic
+    public fun pay2tr(internalKey: XonlyPublicKey, scripts: ScriptTree?): List<ScriptElt> {
```
sstone, line 493: We should also let user provide a merkle root instead of a script tree

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
```diff
@@ -16,45 +16,73 @@
...
+                    leftProof + this.right.hash()
+                } else {
+                    val rightProof = this.right.merkleProof(leafId)
+                    if (rightProof.isNotEmpty()) {
+                        // Our target leaf is in that subtree: we add its sibling to the proof.
+                        rightProof + this.left.hash()
+                    } else {
+                        listOf()
+                    }
+                }
```
sstone, line 85: Nit: it could be refactored into:
```kotlin
    public fun merkleProof(leafId: Int): List<ByteVector32> {
        return when {
            this is Leaf && this.id == leafId -> {
                // We found our leaf: we can now walk up the tree to build the rest of the proof.
                listOf(this.hash())
            }
            this is Branch && this.left.merkleProof(leafId).isNotEmpty() -> {
                // Our target leaf is in that subtree: we add its sibling to the proof.
                this.left.merkleProof(leafId) + this.right.hash()
            }
            this is Branch && this.right.merkleProof(leafId).isNotEmpty() -> {
                // Our target leaf is in that subtree: we add its sibling to the proof.
                this.right.merkleProof(leafId) + this.left.hash()
            }
            else -> listOf()
        }
    }
```

sstone (changes requested):
Nice! Just a few comments and  nits.

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: Why do you think this is necessary? We only have use-cases where we know the whole script tree, so why bother with more complexity before we actually have use-cases for it?

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
t-bast: True, but in the `Branch` case it means we need to compute `merkleProof(leafId)` twice, which isn't very satisfying...

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: Done in https://github.com/ACINQ/bitcoin-kmp/pull/109/commits/e6fb4c054dc24846d2f80711d2696af4b8542262, this one makes sense when building outputs that you may not know how to spend, you may only have access to the merkle root indeed.

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
sstone: Because one of the benefits of MAST is the ability to spend from a specific branch without knowing  the entire script tree but just what is necessary: the script root and a merkle proof (it is very useful to provide a high-level builder that takes a script tree and a specific leaf but as an addition, not a replacement).

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
sstone: Yes but it's just micro-seconds here and the intent is more explicit

t-bast: Fair enough, done in https://github.com/ACINQ/bitcoin-kmp/pull/109/commits/7495aff2d2500d52526c380e5dc1672d8c03777f

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
t-bast: I agree that this is one of the theoretical use-case for MAST, but I'm pretty sure we won't use it in the short term, so I would have rather waited for a use-case to create better API for it. I've still added it in https://github.com/ACINQ/bitcoin-kmp/pull/109/commits/67589c651353f77be75d1592da293c76716b50a0, but I think it's a bad API that is too vague and too easy to misuse. Ideally users of the library should never be directly exposed to the control block, they should have helper functions that let them directly get `ScriptWitness` objects for their use-case.

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
sstone: I've changed this in https://github.com/ACINQ/bitcoin-kmp/pull/111. I don't think that returning a list is the right API here

t-bast: Good idea, the `merkleProof` function is only used to build control blocks, so we can directly return it in the format it expects (and the function is more elegant that way). I took the main change of your PR in https://github.com/ACINQ/bitcoin-kmp/pull/109/commits/5b679b6c6642fa6f6a3573369bf5aa309377d691, and omitted that seemed redundant.

src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -740,7 +741,9 @@ public data class Transaction(
             inputs: List<TxOut>,
             sighashType: Int,
             sigVersion: Int,
-            executionData: Script.ExecutionData = Script.ExecutionData.empty
+            tapleaf: ByteVector32? = null,
```
sstone, line 744: Why split `executionData` into separate fields ? As much as possible, we try to match bitcoin core's implementation choices.

t-bast: Because it's not an `executionData` that is needed here, since it has at least one field of `executionData` that it doesn't use (`validationWeightLeft`). I really believe that `Script.ExecutionData` is an internal detail that should absolutely not be exposed to users of that library (and with that change, we don't expose it anymore). It is mostly used as a mutable variable during script validation, while hashing a transaction is used in other contexts as well.

Bitcoin Core isn't a library, so it doesn't care about the API it exposes on such functions, that's why it's okay-ish to reuse an object that doesn't completely fit in the context. But here this is a function we expose, so we should be more strict about what exactly we expose and avoid exposing unnecessary internal details.

sstone: Ok, but then `ExecutionData`'s visibility should be reduced (probably to `internal`)

t-bast: Yes that would be great! This isn't in the scope of this PR at all though, but several classes inside `Script` (and in other places as well) should be made `internal`, we expose way too much low-level details that callers don't need to know about.

sstone (approved):

PR #111, opened 2024-01-16
Change how taproot merkle proofs are built

sstone:
Returning a single nullable byte array  feels cleaner (and internally this is how such proof are used), remove the ambiguity of returning an empty list, and the need to remove the last element.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
@@ -16,6 +16,7 @@
 package fr.acinq.bitcoin
 
 import fr.acinq.bitcoin.io.ByteArrayOutput
+import kotlin.jvm.JvmStatic
 
 /** Simple binary tree structure containing taproot spending scripts. */
 public sealed class ScriptTree {
@@ -42,6 +43,7 @@ public sealed class ScriptTree {
             BtcSerializer.writeScript(this.script, buffer)
             Crypto.taggedHash(buffer.toByteArray(), "TapLeaf")
         }
+
         is Branch -> {
             val h1 = this.left.hash()
             val h2 = this.right.hash()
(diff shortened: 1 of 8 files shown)


src/commonTest/kotlin/fr/acinq/bitcoin/reference/BIP341TestsCommon.kt
```diff
@@ -101,11 +101,18 @@ class BIP341TestsCommon {
             when (expected["scriptPathControlBlocks"]) {
                 null, JsonNull -> Unit
                 else -> {
+
+                    /** Return the first script leaf with the corresponding id, if any. */
+                    fun findScript(tree: ScriptTree, id: Int): ScriptTree.Leaf? = when (tree) {
+                        is ScriptTree.Leaf -> if (tree.id == id) tree else null
+                        is ScriptTree.Branch -> findScript(tree.left, id) ?: findScript(tree.right, id)
+                    }
```
t-bast, line 109: This is already a method of `ScriptTree`, we don't need to redefine it here?

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
```diff
@@ -50,32 +52,36 @@ public sealed class ScriptTree {
...
+    public fun merkleProof(leaf: Leaf): ByteArray? = merkleProof(this, leaf)
+
+    public companion object {
+        /**
+         * @param tree script tree
+         * @param leafHash target leaf hash
+         * @return a merkle proof for the target leaf hash, or null if it cannot be found in the tree
+         */
+        @JvmStatic
+        public fun merkleProof(tree: ScriptTree, leafHash: ByteVector32): ByteArray? {
```
t-bast, line 74: I don't understand why you put those functions in the companion object, they make more sense applied directly to a `ScriptTree` instance?

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
```diff
@@ -50,32 +52,36 @@ public sealed class ScriptTree {
...
+
+            fun loop(t: ScriptTree, p: ByteArray): ByteArray? = when (t) {
```
t-bast, line 76: Good idea, this is more elegant than the previous solution.

t-bast (commented):
Thanks, I took the main change of this PR directly in https://github.com/ACINQ/bitcoin-kmp/pull/109/commits/5b679b6c6642fa6f6a3573369bf5aa309377d691

src/commonTest/kotlin/fr/acinq/bitcoin/reference/BIP341TestsCommon.kt
sstone: I removed it because it's only used in tests

src/commonMain/kotlin/fr/acinq/bitcoin/ScriptTree.kt
sstone: Here it's both, but the "static" methods you're right could be removed

src/commonTest/kotlin/fr/acinq/bitcoin/reference/BIP341TestsCommon.kt
t-bast: It is meant to be used by applications, to find a specific script leaf in a given tree, that's why I added it in the first place.

sstone: I'm not sure I see the use case there (which does not mean it does not exist), but I don't have a problem with keeping this method in `ScriptTree`

t-bast: I was thinking mostly of minimizing disk storage if we need to store those `ScriptTree`s and restore the leaves that belong to us. I'm not 100% sure we'll need it, since we probably will only use deterministic trees in the short term that we can simply rebuild entirely on-the-fly, but since it's a trivial helper that is at least used for tests I thought it wouldn't hurt to expose it and potentially avoid shipping a new version just for that if we need it later.

sstone:
Some of these changes we added to https://github.com/ACINQ/bitcoin-kmp/pull/109

PR #114, opened 2024-01-25
Add high-level helpers for using Musig2 with Taproot

t-bast:
When using Musig2 for a taproot key path, we can provide simpler helper functions to collaboratively build a shared signature for the spending transaction. Those helper functions hide the low-level details of using an opaque key aggregation cache or signing session. This comes with a small performance penalty, as we recompute the key aggregation cache.

We also document the exposed APIs, import more tests from the official test vectors, and make APIs safe: they should never throw exceptions, except when invalid public keys are provided as inputs (which should be verified by the caller beforehand).

I removed the musig2 prototype as I don't see any valid reason to keep it: it doesn't even have the same API as the official implementation anymore so it's rather confusing, and can be found in the git history if necessary.

This is a PR targeting #107


src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -18,126 +18,150 @@ public data class KeyAggCache(val data: ByteVector) {
...
-         * @param cache an optional key aggregation cache
-         * @return a new (if cache was null) or updated cache, and the aggregated public key
+         * @param publicKeys public keys to aggregate.
+         * @return an opaque key aggregation cache and the aggregated public key.
          */
         @JvmStatic
-        public fun add(pubkeys: List<PublicKey>, cache: KeyAggCache?): Pair<XonlyPublicKey, KeyAggCache> {
-            val localCache = cache?.data?.toByteArray() ?: ByteArray(Secp256k1.MUSIG2_PUBLIC_KEYAGG_CACHE_SIZE)
-            val aggkey = Secp256k1.musigPubkeyAgg(pubkeys.map { it.value.toByteArray() }.toTypedArray(), localCache)
+        public fun create(publicKeys: List<PublicKey>): Pair<XonlyPublicKey, KeyAggCache> {
```
sstone, line 48: This means that we cannot add new public keys to an existing key aggregation cache, which was possible before.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -146,26 +170,126 @@ public data class IndividualNonce(val data: ByteVector) {
...
     public companion object {
+        /**
+         * Aggregate public nonces from all participants of a signing session.
+         * Returns null if one of the nonces provided is invalid.
+         */
         @JvmStatic
-        public fun aggregate(nonces: List<IndividualNonce>): AggregatedNonce {
-            val agg = Secp256k1.musigNonceAgg(nonces.map { it.toByteArray() }.toTypedArray())
-            return AggregatedNonce(agg)
+        public fun aggregate(nonces: List<IndividualNonce>): AggregatedNonce? {
```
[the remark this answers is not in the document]
sstone: It's a generic problem with this library but I don't think that returning `null` is a good option.
The current "best practice" in kotlin is simple to throw exceptions.
Returning `Result` in a public API is messy because it's an inline type, using `arrow` is imho overkill.
Using a custom type as we dif for address parsing is clumsy, we could also use our own `Either` type, which we also define in `lightning-kmp` (this also seems to be a common pattern in kotlin).
I thinkl we should stick to using exceptions, and if not byte the bullet and use this library's `Either` type.

src/commonTest/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2TestsCommon.kt
```diff
@@ -20,20 +17,45 @@ class Musig2TestsCommon {
...
+        require(serialized.size == 97) { "secret nonce from test vector should be serialized using 97 bytes" }
+        // In test vectors, secret nonces are serialized as: <scalar_1> <scalar_2> <compressed_public_key>
+        val compressedPublicKey = PublicKey.parse(serialized.takeLast(33).toByteArray())
+        // We expect secret nonces serialized as: <magic> <scalar_1> <scalar_2> <public_key_x> <public_key_y>
+        // Where we use a different endianness for the public key coordinates than the test vectors.
+        val uncompressedPublicKey = compressedPublicKey.toUncompressedBin()
+        val publicKeyX = uncompressedPublicKey.drop(1).take(32).reversed().toByteArray()
+        val publicKeyY = uncompressedPublicKey.takeLast(32).reversed().toByteArray()
+        val magic = Hex.decode("220EDCF1")
+        return SecretNonce(magic + serialized.take(64) + publicKeyX + publicKeyY)
```
sstone, line 56: This will likely break when we update `secp256k1` through `secp256k1-kmp`. If we really want to add these low-level checks (not sure we need to), they belong in  `secp256k1-kmp`.

sstone (commented):
I'm not convinced that it makes the code that uses musig2 simpler or more compact. For now, this code is encapsulated in specific `swap-in protocol` objects (for both `lightning-kmp` and `eclair`), this is where we can hide musig2 complexity from the rest of the codebase.
Also, the really tricky part imho (same with singing regular transaction inputs) is how to properly set up a musig2 signing session and this requires users to pass the right transaction, index, nonces etc... and it can't really be made simpler.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
t-bast: Yes, it was possible, but not used anywhere. I don't see any scenario where that is useful: why create a key aggregation cache before you have all public keys, since you need to have all of them later? A general rule of library design is to start simple and only introduce more complexity when it's necessary, and here we were really doing the opposite...

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
t-bast: > The current "best practice" in kotlin is simply to throw exceptions.
> I think we should stick to using exceptions.

I strongly disagree with that (where did you see that as a best practice?). You cannot even know if a function throws an exception, which means that callers **will** forget to wrap it in a try/catch, which can cause CVEs. We have good types in the language, we should never throw exceptions, ever, as encoding failure information in the return type is strictly superior and is a best practice.

> byte the bullet and use this library's Either type.

The issue is that it will conflict with `lightning-kmp`'s `Either` type, won't it? So it will be clumsy to use.

> using arrow is imho overkill.

Why is it overkill? We only plan to use the core library, which introduces `Option`, `Try` and `Either`, this should be light enough?

src/commonTest/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2TestsCommon.kt
t-bast: You mean that `secp256k1` may break the format they use for `SecretNonce`s? If so yes, we'll need to update those tests, but this is only used for tests, to make sure we're able to validate the official test vectors. I like validating the test vectors at the `bitcoin-kmp` layer rather than in `secp256k1-kmp`, because they could pass in `secp256k1-kmp`, but `bitcoin-kmp` may use incorrectly the functions exposed, and then we wouldn't have tests to catch it.

t-bast:
> I'm not convinced that it makes the code that uses musig2 simpler or more compact. For now, this code is encapsulated in specific swap-in protocol objects (for both lightning-kmp and eclair), this is where we can hide musig2 complexity from the rest of the codebase.

I really don't understand how you can't be convinced. We're now exposing an API that makes it trivial to spend any output that uses musig2 for its key path with any script tree, without having to know or care about low-level musig2 details (key aggregation cache and opaque session data). That's exactly the abstraction that every use-case of musig2 needs? So instead of duplicating this code in `lightning-kmp` and `eclair`, we now have a single abstraction in `bitcoin-kmp` that we can easily use for both swap-in-potentiam, taproot channels and PTLCs.

The musig2 complexity should be hidden in `bitcoin-kmp` as much as possible, otherwise we're not providing any useful abstraction and multiple users of `bitcoin-kmp` will have to write the same code many times.

> Also, the really tricky part imho (same with singing regular transaction inputs) is how to properly set up a musig2 signing session and this requires users to pass the right transaction, index, nonces etc... and it can't really be made simpler.

I don't understand what you mean: a signer only has to generate his secret nonce, receive public nonces from other participants, and then the existing functions make it trivial to spend the utxos. I believe the swap-in-potentiam unit test showcases that pretty well?

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
sstone: > > The current "best practice" in kotlin is simply to throw exceptions.
> > I think we should stick to using exceptions.
> 
> I strongly disagree with that (where did you see that as a best practice?). You cannot even know if a function throws an exception, which means that callers **will** forget to wrap it in a try/catch, which can cause CVEs. We have good types in the language, we should never throw exceptions, ever, as encoding failure information in the return type is strictly superior and is a best practice.
> 

"best practices" here as in that what most library do since there is not proper built-in type for returning errors, does not mean that it's good.

> > using arrow is imho overkill.
> 
> Why is it overkill? We only plan to use the core library, which introduces `Option`, `Try` and `Either`, this should be light enough?

because we need * very * good support for `ios`, which is often overlooked even in fairly popular libraries (and we'd like to have support for `iosSimulatorArm64` too).

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
sstone: > ... A general rule of library design is to start simple and only introduce more complexity when it's necessary, and here we were really doing the opposite...

No, we keep things simple by exposing the musig2 API as-is and not second-guessing how it could be improved.
We've had very few bugs in this library, and always because we (it's a royal "we" here) we trying to be smarter than bitcoin-core.

src/commonTest/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2TestsCommon.kt
sstone: Verifying that the public nonce is correct should be enough here.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
t-bast: > because we need * very * good support for ios, which is often overlooked even in fairly popular libraries (and we'd like to have support for iosSimulatorArm64 too).

I'd be surprised if arrow didn't have good support for all multiplatform targets, as it contains very simple code that doesn't need any dependency (a bit like our implementation of `Either`).

But even without that, returning a nullable type is IMO much better than throwing exceptions: nobody is interested in the body of those exceptions (they'll only contain low-level secp256k1 information), but it is very important for callers to know which function calls may fail.

src/commonTest/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2TestsCommon.kt
t-bast: It's true for the `secret_nonce_gen` test vectors, but without that function we cannot run the signing test vectors, nor the aggregation test vectors, which are pretty important test vectors to have!

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
sstone: this library does not have any external dependencies, adding `arrow` is imho not a good idea (it means phoenix would depend upon it too). 
We have 3 options:
1. throw exceptions (which is what most low-level methods in this library do)
2. return null in case of error
3. use export this library's `Either` type
4. add custom result types everywhere

I like 1 and 3, 2 could be an intermediary step before we decide on what to do

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -146,26 +170,126 @@ public data class IndividualNonce(val data: ByteVector) {
...
+     * @param scriptTree tapscript tree of the taproot input, if it has script paths.
+     */
+    @JvmStatic
+    public fun aggregateTaprootSignatures(
+        partialSigs: List<ByteVector32>,
+        tx: Transaction,
+        inputIndex: Int,
+        inputs: List<TxOut>,
+        publicKeys: List<PublicKey>,
+        publicNonces: List<IndividualNonce>,
```
sstone, line 289: This means that all public nonces and public keys must be available. This may not be practical, users could have discarded public nonces and just have kept the aggregated nonce. This is what we currently do in https://github.com/ACINQ/lightning-kmp/pull/563.

t-bast: > This is what we currently do in https://github.com/ACINQ/lightning-kmp/pull/563.

But it's much simpler if eclair/lightning-kmp don't have to care about nonce aggregation at all, isn't it? It's simpler to just store the local and remote nonces and then pass them to the signing helper. This gets rid of one "internal detail" of musig2 that higher-level applications don't even need to know about, that's part of encapsulating the complexity inside `bitcoin-kmp`.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
t-bast: 1 is definitely a no-go for me, it's one of the most important best practices IMO to avoid throwing exceptions when it's unnecessary.

I agree that 4 is usually quite annoying, and wouldn't work well in this case since we don't have well-defined errors to return (we would just wrap an opaque error coming from libsecp).

I'm ok with doing 3, but that should be a separate PR: if we do that we should import `Either` and `Try` from `lightning-kmp`, and remove the ones in `lightning-kmp` to avoid type confusion...so I think we should stick to option 2 for now, and later migrate to use `Either` / `Try`. How does that sound?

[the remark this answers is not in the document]
sstone: It sounds good, and returning `null` here is better, I'll add this change to the musig2 PR. I also agree that exporting our `Either` type later should be done later.

src/commonTest/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2TestsCommon.kt
sstone: It should be possible to generate the secret nonces used in these tests (though it does not seem to be documented), I'll investigate.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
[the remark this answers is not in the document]
t-bast: We're now using `Either`.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
t-bast: I have updated `lightning-kmp` to simply keep the local and remote nonce instead of aggregating them in https://github.com/ACINQ/lightning-kmp/pull/591 and it works well.

sstone (approved):
I updated and tested lightning-kmp/eclair/phoenix with this changes and everything works, and it seems that it makes the code in `lightning-kmp` more readable so we can go with this. We'll end up with an API that is a bit more restrictive than the raw musig2 API but we can always revisit this later.

PR #115, opened 2024-01-31
Use Either for error handling

sstone:
We already export an `Either` type, which can be reused by clients of this library.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,7 +12,7 @@ plugins {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 group = "fr.acinq.bitcoin"
-version = "0.17.0-EITHER-SNAPSHOT"
+version = "0.17.0-SNAPSHOT"
 
 repositories {
     google()
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -12,7 +12,7 @@ plugins {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 group = "fr.acinq.bitcoin"
-version = "0.17.0-SNAPSHOT"
+version = "0.17.0-EITHER-SNAPSHOT"
 
 repositories {
     google()
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
@@ -27,14 +27,45 @@ public fun <T> List<T>.updated(i: Int, t: T): List<T> = when (i) {
     else -> this.take(i) + t + this.drop(i + 1)
 }
 
-public sealed class BitcoinException(message: String, cause: Throwable?) : Exception(message, cause) {
-    public data object InvalidChainHash : BitcoinException("invalid chain hash", null)
-    public data object ChainHashMismatch : BitcoinException("chain hash mismatch", null)
-    public data object InvalidScript : BitcoinException("invalid script", null)
-    public data object InvalidAddress : BitcoinException("invalid address", null)
-    public data object InvalidBech32Address : BitcoinException("invalid address", null)
-    public data class InvalidWitnessVersion(val version: Int) : BitcoinException("invalid witness version $version", null)
-    public data class GenericError(override val message: String, override val cause: Throwable?) : BitcoinException(message, cause)
+public sealed class BitcoinError {
+    public abstract val message: String
+    public abstract val cause: Throwable?
+    override fun toString(): String = when (cause) {
+        null -> message
+        else -> "$message: ${cause?.message}"
+    }
+
+    public data object InvalidChainHash : BitcoinError() {
+        override val message: String = "invalid chain hash"
+        override val cause: Throwable? = null
+    }
+
+    public data object ChainHashMismatch : BitcoinError() {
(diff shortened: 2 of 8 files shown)


build.gradle.kts
```diff
@@ -12,7 +12,7 @@ plugins {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 group = "fr.acinq.bitcoin"
-version = "0.17.0-SNAPSHOT"
+version = "0.17.0-EITHER-SNAPSHOT"
```
t-bast, line 15: I think we should revert that change and directly integrate this PR with the snapshot version.

src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -26,63 +27,14 @@ public fun <T> List<T>.updated(i: Int, t: T): List<T> = when (i) {
...
-    }
+public sealed class BitcoinException(message: String, cause: Throwable?) : Exception(message, cause) {
+    public data object InvalidChainHash : BitcoinException("invalid chain hash", null)
+    public data object ChainHashMismatch : BitcoinException("chain hash mismatch", null)
+    public data object InvalidScript : BitcoinException("invalid script", null)
+    public data object InvalidAddress : BitcoinException("invalid address", null)
+    public data object InvalidBech32Address : BitcoinException("invalid address", null)
+    public data class InvalidWitnessVersion(val version: Int) : BitcoinException("invalid witness version $version", null)
+    public data class GenericError(override val message: String, override val cause: Throwable?) : BitcoinException(message, cause)
 }
```
t-bast, line 38: There is no reason for this class to be an `Exception`, it's more verbose without it but it's cleaner (functional programming FTW). It's also more efficient: some months ago Thomas investigated the cost of using exceptions because he was skeptical that they were really bad, and his benchmarks showed that it's actually much more costly than we'd think, because a stack trace is captured when the exception is _created_, not just when it's _thrown_.

```suggestion
public sealed class BitcoinError {
    public abstract val message: String
    public abstract val cause: Throwable?

    override fun toString(): String = when (cause) {
        null -> message
        else -> "$message: ${cause?.message}"
    }

    public data object InvalidChainHash : BitcoinError() {
        override val message: String = "invalid chain hash"
        override val cause: Throwable? = null
    }

    public data object ChainHashMismatch : BitcoinError() {
        override val message: String = "chain hash mismatch"
        override val cause: Throwable? = null
    }

    public data object InvalidScript : BitcoinError() {
        override val message: String = "invalid script"
        override val cause: Throwable? = null
    }

    public data object InvalidAddress : BitcoinError() {
        override val message: String = "invalid address"
        override val cause: Throwable? = null
    }

    public data object InvalidBech32Address : BitcoinError() {
        override val message: String = "invalid bech32 address"
        override val cause: Throwable? = null
    }

    public data class InvalidWitnessVersion(val version: Int) : BitcoinError() {
        override val message: String = "invalid witness version $version"
        override val cause: Throwable? = null
    }

    public data class GenericError(override val message: String, override val cause: Throwable?) : BitcoinError()
}
```

t-bast (commented):
I guess the goal is to delete the corresponding class from `lightning-kmp` when we update it to use this version of `bitcoin-kmp`, right? In that case we should also import the `Try` implementation, this way we remove it as well from `lightning-kmp`, otherwise it would be weird to keep one in `lightning-kmp` and the other in `bitcoin-kmp`?

build.gradle.kts
sstone: It makes it much easier to test locally when there are many different branches, and it also makes it possible to publish snapshots that can be used in other PRs (other CI would never work), I'll update the version once the PR has been approved.

sstone:
> I guess the goal is to delete the corresponding class from `lightning-kmp` when we update it to use this version of `bitcoin-kmp`, right? In that case we should also import the `Try` implementation, this way we remove it as well from `lightning-kmp`, otherwise it would be weird to keep one in `lightning-kmp` and the other in `bitcoin-kmp`?

Done in https://github.com/ACINQ/bitcoin-kmp/pull/115/commits/7af325c26edf35a30d7fb2c8bd117d337c8d092b, though I believe that we may not need it: API methods in `lightning-kmp` should use `Either`, and internal methods could just use kotlin's built-in `Result` type

t-bast (commented):
LGTM, I believe we can revert the version change and merge this :+1:

t-bast:
> Done in https://github.com/ACINQ/bitcoin-kmp/commit/7af325c26edf35a30d7fb2c8bd117d337c8d092b, though I believe that we may not need it: API methods in lightning-kmp should use Either, and internal methods could just use kotlin's built-in Result type

I agree with that, but inside `lightning-kmp` we do use `Try` in a few places, and it will be cleaner to use the one exported by `bitcoin-kmp`. Then maybe we can stop using `Try` altogether from `ligthning-kmp` if it doesn't make sense, but we can do that in a next step, no need to do it right now!

sstone:
> LGTM, I believe we can revert the version change and merge this 👍

done in https://github.com/ACINQ/bitcoin-kmp/pull/115/commits/e3ba13542eb42a7e93c2ca572767967179955986

t-bast (approved):

PR #117, opened 2024-02-13
Add Chain type

sstone:


src/commonMain/kotlin/fr/acinq/bitcoin/Bitcoin.kt
```diff
@@ -234,4 +234,22 @@ public object Bitcoin {
...
+        public object Mainnet : Chain("Mainnet", Block.LivenetGenesisBlock)
+
+        public fun isMainnet(): Boolean = this is Mainnet
+
+        public fun isTestnet(): Boolean = this is Testnet
+
+        public val chainHash: BlockHash get() = genesis.hash
+
+        override fun toString(): String = name
+    }
```
pm47, line 254: Looks better and more readable without a few line breaks:

```suggestion
    public sealed class Chain(public val name: String, private val genesis: Block) {
        public object Regtest : Chain("Regtest", Block.RegtestGenesisBlock)
        public object Testnet : Chain("Testnet", Block.TestnetGenesisBlock)
        public object Signet : Chain("Signet", Block.SignetGenesisBlock)
        public object Mainnet : Chain("Mainnet", Block.LivenetGenesisBlock)

        public fun isMainnet(): Boolean = this is Mainnet
        public fun isTestnet(): Boolean = this is Testnet

        public val chainHash: BlockHash get() = genesis.hash

        override fun toString(): String = name
    }
```

Why the redundant `public` visibility modifiers?

sstone: done  in https://github.com/ACINQ/bitcoin-kmp/pull/117/commits/600fc840338e652c40ec4b5165c9532b67d63529

The `public` modifiers are needed because this library is built in  [explicit api mode ](https://kotlinlang.org/docs/whatsnew14.html#explicit-api-mode-for-library-authors)

t-bast (approved):

pm47 (approved):

PR #127, opened 2024-05-27
Update musig2 api

sstone:
To use taproot in lightning (to implement simple taproot channels for example) we need:
- to create musig2 sessions to verify partial signatures when we cannot yet aggregate them into a full signature.
- to read/write tapscript script trees


sstone:
superseded by https://github.com/ACINQ/bitcoin-kmp/pull/129

PR #130, opened 2024-08-12
Add support for testnet4

sstone:
See https://github.com/bitcoin/bitcoin/pull/29775.
It basically adds a new genesis block the testnet4  (addresses uses the same prefixes/encoding as testnet3),


sstone:
Replaced by https://github.com/ACINQ/bitcoin-kmp/pull/136

PR #134, opened 2024-08-21
Improvements and bug fixes for BIP 371 support

t-bast:
This commit fixes my comments made on #128 and adds a few improvements:

- fix a bug in the index used for previous inputs
- use `InvalidTxOutput` for invalid taproot output fields
- add test vectors from BIP 371
- reformat to match the rest of the file

Note that I updated the "build and sign a BIP86 psbt" test to include a non-segwit input, this way it fails without the fix to the input index.

Note: this is a PR targetting the #128 branch.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
@@ -17,7 +17,6 @@
 package fr.acinq.bitcoin.psbt
 
 import fr.acinq.bitcoin.*
-import fr.acinq.bitcoin.Transaction.Companion.hashForSigningSchnorr
 import fr.acinq.bitcoin.crypto.Pack
 import fr.acinq.bitcoin.io.ByteArrayInput
 import fr.acinq.bitcoin.io.ByteArrayOutput
@@ -96,7 +95,6 @@ public data class Psbt(@JvmField val global: Global, @JvmField val inputs: List<
                 taprootInternalKey ?: input.taprootInternalKey,
                 input.unknown
             )
-
             is Input.WitnessInput.PartiallySignedWitnessInput -> input.copy(
                 txOut = txOut,
                 redeemScript = redeemScript ?: input.redeemScript,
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
```diff
@@ -96,7 +95,6 @@ public data class Psbt(@JvmField val global: Global, @JvmField val inputs: List<
                 taprootInternalKey ?: input.taprootInternalKey,
                 input.unknown
             )
-
```
sstone, line 99: nit: Idea keeps re-adding this line breaks

src/commonTest/kotlin/fr/acinq/bitcoin/psbt/PsbtTestsCommon.kt
```diff
@@ -159,6 +157,36 @@ class PsbtTestsCommon {
                 )
             )
         )
+        // PSBT with invalid taproot internal key (incorrectly serialized as compressed DER)
+        assertEquals(
+            Either.Left(ParseFailure.InvalidTxInput("taproot internal key entry must have a 32 bytes value")),
+            Psbt.read(ByteVector("70736274ff010071020000000127744ababf3027fe0d6cf23a96eee2efb188ef52301954585883e69b6624b2420000000000ffffffff02787c01000000000016001483a7e34bd99ff03a4962ef8a1a101bb295461ece606b042a010000001600147ac369df1b20e033d6116623957b0ac49f3c52e8000000000001012b00f2052a010000002251205a2c2cf5b52cf31f83ad2e8da63ff03183ecd8f609c7510ae8a48e03910a075701172102fe349064c98d6e2a853fa3c9b12bd8b304a19c195c60efa7ee2393046d3fa232000000")),
+        )
```
sstone, line 164: nit: using the base64 version makes it easier to test with bitcoin-core

sstone (approved):
Just a nits but it can be merged as is. Thanks!

PR #140, opened 2024-09-17
Upgrade to Kotlin 2.0.20

sstone:
PR will be marked as ready for review once `secp256k1-kmp` 0.16.0 has been released and the build updated to use it instead of the SNAPSHOT version.


sstone:
Close as upgrading to Kotlin 2.0 causes issues with Phoenix (see https://github.com/ACINQ/phoenix/pull/628)

PR #143, opened 2025-02-03
Verify dependency checksums

sstone:


pm47:
Why are you going with SHA256 here, but went with SHA512 in eclair?

sstone:
> Why are you going with SHA256 here, but went with SHA512 in eclair?

No reason I'll switch to SHA512 to be consistent with eclair.

sstone:
> Same comment as eclair, can you document how you initially generated those checksums, and the process to update them easily?

Done in https://github.com/ACINQ/bitcoin-kmp/pull/143/commits/b1bed42fee92ee1c50096e36e6dfac70771c7a61. Checksums were initially generated with gradle and verified on 3 different machines (excluding github CI).

PR #155, opened 2025-07-03
Set compiler option to target java 1.8

sstone:
Setting `jvmTarget` to target java 1.8 is not enough, Android apps may fail with an error like:

```
java.lang.NoSuchMethodError: No interface method removeFirst()Ljava/lang/Object; in class Ljava/util/List; or its super classes (declaration of 'java.util.List' appears in /system/framework/core-oj.jar)
``` 

 see https://jakewharton.com/kotlins-jdk-release-compatibility-flag/ and https://youtrack.jetbrains.com/issue/KT-49746/.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,12 +13,11 @@ plugins {
 val currentOs = org.gradle.internal.os.OperatingSystem.current()
 
 group = "fr.acinq.bitcoin"
-version = "0.25.0"
+version = "0.25.1-SNAPSHOT"
 
 repositories {
     google()
     mavenCentral()
-    maven(url="https://oss.sonatype.org/content/repositories/snapshots/")
 }
 
 kotlin {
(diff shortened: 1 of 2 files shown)


dpad85 (approved):
Tested on a device (android 15) and a simulator (android 13) and the error only happens on the simulator, not on the device.

Example on a simulator:

```
2025-07-07 11:26:12.174 f.a.l.c.s.ChannelState            E   - error on message CommitSig
at fr.acinq.lightning.channel.InvalidCommitmentSignature: invalid commitment signature: txId=e68c3af090e31041989d181bb13b260fe232ab5b778222ef9217404a3cd18f95
at fr.acinq.lightning.channel.LocalCommit$Companion.fromCommitSig(Commitments.kt:123)
at fr.acinq.lightning.channel.Commitment.receiveCommit(Commitments.kt:531)
```

This error closes the channel. After updating with this PR, the issue does not happen anymore.

PR #174, opened 2026-03-17
Upgrade to secp256k1-kmp 0.23.0

sstone:
- Upgrade to kotlin 2.3.10
- Use compact signatures with secp256k1-kmp
- Upgrade reference test vectors (taken from  bitoin core at d9c7364ac56781a16c7224b2c7a6db9db97f17d8)

PR currently uses secp256k1-kmp 0.23.0-SNAPSHOT which has been deployed to  https://central.sonatype.com/repository/maven-snapshots and should not change anymore, I will update to 0.23.0 once it's released.

EDIT: PR has been updated to use secp256k1-kmp 0.23.0


src/commonTest/kotlin/fr/acinq/bitcoin/MnemonicCodeTestsCommon.kt
```diff
@@ -18,6 +18,7 @@ package fr.acinq.bitcoin
 
 import fr.acinq.bitcoin.MnemonicCode.toMnemonics
 import fr.acinq.bitcoin.MnemonicCode.toSeed
+import fr.acinq.bitcoin.utils.runTrying
```
t-bast: nit: revert?

src/commonMain/kotlin/fr/acinq/bitcoin/Crypto.kt
```diff
@@ -340,35 +328,83 @@ public object Crypto {
         return true
     }
 
+    /**
+     * Decode a DER-encoded signature and return a compact signature. Allows for some violations of DER-encoding rules.
+     * @param derSignature DER-encoded signature
+     * @return a compact 64 bytes signature
+     */
```
t-bast, line 319: Can you provide a better explanation for *why* we need that function? If we remove it and use secp256k1's `der2compact` we indeed have some reference test vector failing, but it's unclear to me why we need this lax decoding.

Also, it's unclear where the implementation comes from: if it was inspired from bitcoind's implementation, can you point me to where that is done? It is important to ensure that we perfectly match them here, otherwise we may accept signatures that they reject or the other way around, which would be a very dangerous bug.

sstone: added comments to trace the source for this method in https://github.com/ACINQ/bitcoin-kmp/pull/174/changes/4186a2ad44946b2f39f4b4717d4b571e5fc70e3a

t-bast (approved):
LGTM, this clear up confusion about the bitcoin-specific der-encoding 👍

PR #177, opened 2026-04-07
Fix parser, signing, and MuSig2 issues

jmecom:
## Summary

This PR fixes five security issues in `bitcoin-kmp`. The bugs were all in malformed-input handling or signer safety boundaries: oversized varints could be truncated during deserialization, malformed DER length bytes could crash ECDSA verification, MuSig2 secret nonces could be reused after signing, P2WPKH PSBT signing trusted attacker-supplied `witnessScript`, and taproot `SIGHASH_SINGLE` could dereference an out-of-bounds output index.

None of these fixes should change valid Bitcoin encodings or normal signing flows. The changes harden the library against malformed or malicious inputs and tighten signer behavior so the library matches the security expectations of callers more closely.

### 1. Varint truncation in `BtcSerializer`
`BtcSerializer` previously converted `ULong` lengths and counts to `Int` with unchecked `.toInt()` calls. That allowed attacker-controlled varints to wrap to negative or small positive values, which could:
- bypass length checks
- crash on allocation
- misparse byte streams by consuming only a prefix of attacker-controlled data
- bypass `maxElement` limits for collections

This PR adds explicit bounds checks before every relevant narrowing conversion and rejects lengths or counts that exceed `Int.MAX_VALUE`.

### 2. Signed-byte DER parsing crash in `Crypto`
ECDSA DER parsing used signed `Byte` values directly in length and index arithmetic. Length bytes with the high bit set, including BER long-form lengths, could produce negative offsets and uncaught indexing errors during verification.

This PR masks DER length bytes to unsigned values and makes the lax decoder return an invalid signature result instead of crashing on malformed length encodings.

### 3. MuSig2 `SecretNonce` reuse in `Musig2`
`SecretNonce` was a `data class`, and signing passed a copied buffer into libsecp256k1. That meant native zeroization only affected the copy, while the original `SecretNonce` remained reusable. A caller mistake that reused the same nonce across sessions could expose the signer’s private key.

This PR makes `SecretNonce` single-use:
- signing now consumes the actual backing nonce buffer
- a second signing attempt with the same nonce fails
- `SecretNonce` is no longer a `data class`, so it no longer exposes `copy()` semantics for nonce duplication

### 4. P2WPKH PSBT script-code misbinding in `Psbt`
For P2WPKH inputs, the signer previously used `input.witnessScript` as the script code when present. That let a malicious PSBT coordinator substitute attacker-chosen script code and obtain a signature for a different segwit v0 spending context.

This PR fixes the P2WPKH signing path so it always derives script code from the actual witness program:
- native P2WPKH now signs `OP_DUP OP_HASH160 <20-byte program> OP_EQUALVERIFY OP_CHECKSIG`
- nested P2SH-P2WPKH now derives the same script code from the redeem script’s witness program
- spec-compliant P2WPKH PSBTs no longer require `witnessScript`

### 5. Taproot `SIGHASH_SINGLE` out-of-bounds access in `Transaction`
The taproot schnorr sighash path accessed `txOut[inputIndex]` without checking whether a corresponding output existed. For malformed transactions using `SIGHASH_SINGLE`, that caused an uncaught `IndexOutOfBoundsException`.

This PR adds an explicit guard so the failure is handled intentionally instead of reaching an out-of-bounds array access.

## Implementation notes

The main code changes are in:
- `BtcSerializer.kt`
- `Crypto.kt`
- `crypto/musig2/Musig2.kt`
- `psbt/Psbt.kt`
- `Transaction.kt`

The behavior of valid transactions, signatures, and PSBTs is unchanged. The changes only affect malformed or unsafe cases.

## Verification

Added regression coverage for each fix:
- oversized and truncating varints are rejected
- malformed DER/BER length bytes no longer crash verification helpers
- MuSig2 secret nonces cannot be reused
- P2WPKH PSBT signing derives script code from the witness program and works without `witnessScript`
- malformed taproot `SIGHASH_SINGLE` no longer reaches out-of-bounds output access

Verified with:

```bash
./gradlew jvmTest
```

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
@@ -68,17 +68,9 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
     /**
      * @param secretNonce signer's secret nonce (see [SecretNonce.generate]).
      * @param privateKey signer's private key.
-     * @return a musig2 partial signature.
-     */
-    public fun sign(secretNonce: SecretNonce, privateKey: PrivateKey): ByteVector32 =
-        signEither(secretNonce, privateKey).getOrElse { throw it }
-
-    /**
-     * @param secretNonce signer's secret nonce (see [SecretNonce.generate]).
-     * @param privateKey signer's private key.
      * @return a musig2 partial signature, or an error if the nonce has already been used or signing fails.
      */
-    public fun signEither(secretNonce: SecretNonce, privateKey: PrivateKey): Either<Throwable, ByteVector32> =
+    public fun sign(secretNonce: SecretNonce, privateKey: PrivateKey): Either<Throwable, ByteVector32> =
         secretNonce.consume { nonce ->
             Secp256k1.musigPartialSign(nonce, privateKey.value.toByteArray(), keyAggCache.toByteArray(), this.toByteArray()).byteVector32()
         }
@@ -137,8 +129,7 @@ public class SecretNonce private constructor(bytes: ByteArray, offset: Int, size
 
     internal val data: ByteArray = bytes.copyOfRange(offset, offset + size)
 
-    public constructor(bin: ByteArray) : this(bin, 0, bin.size)
-    public constructor(hex: String) : this(Hex.decode(hex))
+    internal constructor(bin: ByteArray) : this(bin, 0, bin.size)
 
     override fun toString(): String = "<secret_nonce>"
 
@@ -381,7 +372,7 @@ public object Musig2 {
         publicNonces: List<IndividualNonce>,
         scriptTree: ScriptTree?
     ): Either<Throwable, ByteVector32> {
-        return taprootSession(tx, inputIndex, inputs, publicKeys, publicNonces, scriptTree).flatMap { it.signEither(secretNonce, privateKey) }
+        return taprootSession(tx, inputIndex, inputs, publicKeys, publicNonces, scriptTree).flatMap { it.sign(secretNonce, privateKey) }
     }
 
     /**
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
@@ -71,7 +71,7 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
      * @return a musig2 partial signature.
      */
     public fun sign(secretNonce: SecretNonce, privateKey: PrivateKey): ByteVector32 {
-        return Secp256k1.musigPartialSign(secretNonce.data.toByteArray(), privateKey.value.toByteArray(), keyAggCache.toByteArray(), this.toByteArray()).byteVector32()
+        return Secp256k1.musigPartialSign(secretNonce.consume(), privateKey.value.toByteArray(), keyAggCache.toByteArray(), this.toByteArray()).byteVector32()
     }
 
     /**
@@ -119,15 +119,31 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
  * Musig2 secret nonce, that should be treated as a private opaque blob.
  * This nonce must never be persisted or reused across signing sessions.
  */
-public data class SecretNonce(internal val data: ByteVector) {
-    public constructor(bin: ByteArray) : this(bin.byteVector())
-    public constructor(hex: String) : this(Hex.decode(hex))
+public class SecretNonce private constructor(internal val data: ByteArray, @Suppress("UNUSED_PARAMETER") copied: Boolean) {
+    public constructor(bin: ByteArray) : this(bin.copyOf(), true)
+    public constructor(hex: String) : this(Hex.decode(hex), true)
 
     init {
-        require(data.size() == Secp256k1.MUSIG2_SECRET_NONCE_SIZE) { "musig2 secret nonce must be ${Secp256k1.MUSIG2_SECRET_NONCE_SIZE} bytes" }
+        require(data.size == Secp256k1.MUSIG2_SECRET_NONCE_SIZE) { "musig2 secret nonce must be ${Secp256k1.MUSIG2_SECRET_NONCE_SIZE} bytes" }
     }
 
     override fun toString(): String = "<secret_nonce>"
+
+    private var consumed: Boolean = false
+
+    internal fun consume(): ByteArray {
+        check(!consumed) { "secret nonce has already been used" }
+        consumed = true
+        return data
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is SecretNonce) return false
+        return data.contentEquals(other.data)
+    }
+
+    override fun hashCode(): Int = data.contentHashCode()
 
     public companion object {
         /**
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -119,16 +119,32 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
  * Musig2 secret nonce, that should be treated as a private opaque blob.
  * This nonce must never be persisted or reused across signing sessions.
  */
-public data class SecretNonce(internal val data: ByteVector) {
-    public constructor(bin: ByteArray) : this(bin.byteVector())
-    public constructor(hex: String) : this(Hex.decode(hex))
+public class SecretNonce private constructor(internal val data: ByteArray, @Suppress("UNUSED_PARAMETER") copied: Boolean) {
```
t-bast: Can you explain why you're introducing this `copied` boolean? Since it's unused, it's hard to figure out. The two constructors from a `ByteArray` and a `String` are only meant to be used in tests (to read test vectors), application code must always use the static `generateXXX` functions.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -119,16 +119,32 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
...
     }
 
     override fun toString(): String = "<secret_nonce>"
 
+    private var consumed: Boolean = false
+
+    internal fun consume(): ByteArray {
+        check(!consumed) { "secret nonce has already been used" }
+        consumed = true
+        return data
```
t-bast: Have you considered zero-ing out nonce data after use? I don't know if it's effectively useful, since we'd rely on GC to actually collect the array, so I'm not sure if it's worth the additional code, but I'm just curious.

I think the acceptable way to do it would be to return a copy of `data` in this function (which would be GC-ed after use) and do `data.fill(0)` to clear the internal array. Again, I'm not entirely sure this is actually useful.

t-bast (commented):
Thanks, the changes look good! We wanted to use Kotlin's unsigned types to clean up some of the boundaries checks, but it wasn't available when we started, and the migration isn't that easy to do now that we have running code (see https://github.com/ACINQ/lightning-kmp/issues/160).

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
jmecom: I looked into zeroing the nonce after use, but I don’t think it really improves security. The important property is preventing nonce reuse, which we enforce by making SecretNonce single-use. Zeroing the Kotlin-side array would only clear one managed copy, while the nonce already passes through other managed/native buffers during signing, so it doesn’t provide a strong guarantee.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
jmecom: Removed, yeah it didn't make sense

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
t-bast: Agreed!

t-bast (commented):
LGTM 👍 . @sstone will do another pass when we're ready for another release, probably after the next `eclair` and Phoenix releases.

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -118,17 +118,37 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
...
+    internal val data: ByteArray = bytes.copyOfRange(offset, offset + size)
+
+    public constructor(bin: ByteArray) : this(bin, 0, bin.size)
+    public constructor(hex: String) : this(Hex.decode(hex))
+
     override fun toString(): String = "<secret_nonce>"
 
+    private var consumed: Boolean = false
+
+    internal fun consume(): ByteArray {
```
sstone: I'd rather use a more explicit API that forces users to deal with errors, like:
```kotlin
    public fun <T> consume(block: (ByteArray) -> T): Either<Throwable, T> {
        if (used) return Either.Left(IllegalStateException("secret nonce has already been used"))
        used = true
        try {
            return Either.Right(block(data))
        }
        catch(t: Throwable) {
            return Either.Left(t)
        }
    }
```

sstone (changes requested):
Nice! The changes in `SecretNonce` will have an impact on our downstream projects. Some tests will break (because we re-use nonces to sign with invalid data for example) and maybe some of our workflows too(like the simple close workflow where we compute the same local signatures twice). 
=> I'd like to review what it means for these projects before we merge this

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -70,9 +70,18 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
      * @param privateKey signer's private key.
      * @return a musig2 partial signature.
      */
-    public fun sign(secretNonce: SecretNonce, privateKey: PrivateKey): ByteVector32 {
-        return Secp256k1.musigPartialSign(secretNonce.data.toByteArray(), privateKey.value.toByteArray(), keyAggCache.toByteArray(), this.toByteArray()).byteVector32()
-    }
+    public fun sign(secretNonce: SecretNonce, privateKey: PrivateKey): ByteVector32 =
```
sstone: We should just update the API to return `Either`

src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -118,17 +127,41 @@ public data class Session(private val data: ByteVector, private val keyAggCache:
...
     init {
-        require(data.size() == Secp256k1.MUSIG2_SECRET_NONCE_SIZE) { "musig2 secret nonce must be ${Secp256k1.MUSIG2_SECRET_NONCE_SIZE} bytes" }
+        require(size == Secp256k1.MUSIG2_SECRET_NONCE_SIZE) { "musig2 secret nonce must be ${Secp256k1.MUSIG2_SECRET_NONCE_SIZE} bytes" }
+        require(offset >= 0 && offset + size <= bytes.size) { "invalid secret nonce slice" }
     }
 
+    internal val data: ByteArray = bytes.copyOfRange(offset, offset + size)
+
+    public constructor(bin: ByteArray) : this(bin, 0, bin.size)
+    public constructor(hex: String) : this(Hex.decode(hex))
```
sstone: We just need one of them, and just for testing:
```suggestion
    internal constructor(bin: ByteArray) : this(bin, 0, bin.size)
```

sstone (changes requested):
LGTM, just a few nits. I checked that it can be used in  https://github.com/ACINQ/eclair and https://github.com/ACINQ/lightning-kmp with a few simple changes (https://github.com/ACINQ/eclair/pull/3289 and https://github.com/ACINQ/lightning-kmp/pull/856 which should be merged before this PR) .

sstone (approved):

t-bast (approved):
LGTM 👍

PR #178, opened 2026-04-23
Add taproot and musig2 helpers for taproot gossip support

t-bast:
We add several helpers that become useful when implementing support for taproot gossip to verify channel announcements: see https://github.com/ellemouton/bolts/blob/bolt-taproot-gossip/bolt-taproot-gossip.md#taproot-channel-proof-and-verification for more details.

Each commit adds separate helpers:

- a helper function to extract the output key of a p2tr output
- high-level musig2 functions to sign arbitrary messages

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
@@ -345,7 +345,7 @@ public object Musig2 {
      * @return a partial signature, or an error if the nonce has already been used or session creation/signing fails.
      */
     @JvmStatic
-    public fun signMessage(privateKey: PrivateKey, secretNonce: SecretNonce, msg: ByteVector32, publicKeys: List<PublicKey>, publicNonces: List<IndividualNonce>): Either<Throwable, ByteVector32> {
+    public fun sign(privateKey: PrivateKey, secretNonce: SecretNonce, msg: ByteVector32, publicKeys: List<PublicKey>, publicNonces: List<IndividualNonce>): Either<Throwable, ByteVector32> {
         return signingSession(msg, publicKeys, publicNonces).flatMap { session ->
             session.sign(secretNonce, privateKey)
         }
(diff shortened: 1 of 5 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/crypto/musig2/Musig2.kt
```diff
@@ -326,6 +326,62 @@ public object Musig2 {
...
+     *
+     * @param privateKey private key of the signing participant.
+     * @param secretNonce secret nonce of the signing participant.
+     * @param msg message that should be signed.
+     * @param publicKeys public keys of all participants of the musig2 session: callers must verify that all public keys are valid.
+     * @param publicNonces public nonces of all participants of the musig2 session.
+     * @return a partial signature, or an error if the nonce has already been used or session creation/signing fails.
+     */
+    @JvmStatic
+    public fun signMessage(privateKey: PrivateKey, secretNonce: SecretNonce, msg: ByteVector32, publicKeys: List<PublicKey>, publicNonces: List<IndividualNonce>): Either<Throwable, ByteVector32> {
```
sstone: nit: why not just call it `sign()` ?

t-bast: Because we've been explicit in the other signing helper (`signTaprootInput`), so I thought it would make sense to be more precise here as well than just `sign`? But I can change it if you think that `sign` makes more sense!

sstone: `signTaprootInput` is very specialised so it's ok to have a specific name, but here it's a fairly generic signing method, and the corresponding verification method  is simply called `verify()`.

t-bast: Sounds reasonable, done in https://github.com/ACINQ/bitcoin-kmp/pull/178/commits/5da621f3cad59cb14ed6d6fb8bf02a43ee286ca3

sstone (approved):

PR #179, opened 2026-05-05
fix: Generate BIP32 key fingerprints as 4 byte hex strings

shubertm:
@sstone @t-bast 

Currently the function `fingerprint` return `Long` and somehow from byte level computation it returns invalid fingerprints for some keys. So lucky, this can be verified with BIP39 test vectors. 
You can reproduce this by running the tests in `BIP32KeyFingerprintTestsCommon`

The provided fix is in function `keyFingerprint` which returns the fingerprint as an exactly 4 byte hex string, it can be converted to `Long` by `toLong` with radix `16` for any usages that require `Long` fingerprints.

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
@@ -103,6 +103,11 @@ public object DeterministicWallet {
         public fun fingerprint(): Long = extendedPublicKey.fingerprint()
 
         /**
+         * @return the fingerprint for this private key as a 4 byte hex string
+         */
+        public fun keyFingerprint(): String = extendedPublicKey.keyFingerprint()
+
+        /**
          * We avoid accidentally logging extended private keys.
          * You should use an explicit method if you want to convert the extended private key to a string representation.
          */
@@ -164,7 +169,7 @@ public object DeterministicWallet {
                 chaincode = IR.byteVector32(),
                 depth = depth + 1,
                 path = path.derive(index),
-                parent = fingerprint()
+                parent = keyFingerprint().toLong(16)
             )
         }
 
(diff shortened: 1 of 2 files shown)


sstone:
Thanks, I see what the problem is: we use a signed long (64 bits) to represent a 32 bits fingerprints, when it's negative you get an extra 0xFFFFFF... 
I will fix this, meanwhile you can use `.fingerprint() and 0xFFFFFFFFL` which will give you valid results.

shubertm:
> Thanks, I see what the problem is: we use a signed long (64 bits) to represent a 32 bits fingerprints, when it's negative you get an extra 0xFFFFFF... I will fix this, meanwhile you can use `.fingerprint() and 0xFFFFFFFFL` which will give you valid results.

It would also be great to have the SDK return the fingerprint just as a hex.

sstone:
> > Thanks, I see what the problem is: we use a signed long (64 bits) to represent a 32 bits fingerprints, when it's negative you get an extra 0xFFFFFF... I will fix this, meanwhile you can use `.fingerprint() and 0xFFFFFFFFL` which will give you valid results.
> 
> It would also be great to have the SDK return the fingerprint just as a hex.

How to convert a 32 bits value (which is how a fingerprint is defined) to a string  should be up to the user.

shubertm:
> > > Thanks, I see what the problem is: we use a signed long (64 bits) to represent a 32 bits fingerprints, when it's negative you get an extra 0xFFFFFF... I will fix this, meanwhile you can use `.fingerprint() and 0xFFFFFFFFL` which will give you valid results.
> > 
> > 
> > It would also be great to have the SDK return the fingerprint just as a hex.
> 
> How to convert a 32 bits value (which is how a fingerprint is defined) to a string should be up to the user.

Alright, I guess we can close this PR, right?

sstone:
Closing, I've opened an issue for this here:  https://github.com/ACINQ/bitcoin-kmp/issues/180, thanks for the report!

PR #181, opened 2026-05-07
Fix bip32 fingerprint

### LINKED ISSUE #180: BIP32 fingerprint() method result need to be truncated to 32 bits
In some cases, method `fingerprint()` in `ExtendedPrivateKey/ExtendedPublicKey`  will return a result that needs to be truncated to 32 bits.
You can work around this problem by doing the truncation yourself::
```kotlin
val fingerprint = key.fingerprint() and 0xFFFFFFFFL // truncate to 32 bits
``` 

reported by @shubertm

sstone:
Fixes issue #180

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
@@ -174,7 +174,7 @@ public object DeterministicWallet {
 
         public fun derivePublicKey(keyPath: String): ExtendedPublicKey = derivePublicKey(KeyPath.fromPath(keyPath))
 
-        public fun fingerprint(): Long = Pack.int32LE(ByteArrayInput(Crypto.hash160(publickeybytes).take(4).reversed().toByteArray())).toLong()
+        public fun fingerprint(): Long = Pack.int32BE(Crypto.hash160(publickeybytes)).toLong() and 0xFFFFFFFFL
 
         public fun encode(prefix: Int): String {
             val out = ByteArrayOutput()
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
@@ -211,21 +211,6 @@ public object DeterministicWallet {
         }
     }
 
-    @JvmStatic
-    public fun encode(input: ExtendedPrivateKey, testnet: Boolean): String = encode(input, if (testnet) tprv else xprv)
-
-    @JvmStatic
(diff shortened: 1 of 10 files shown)


src/commonMain/kotlin/fr/acinq/bitcoin/DeterministicWallet.kt
```diff
@@ -174,7 +174,7 @@ public object DeterministicWallet {
 
         public fun derivePublicKey(keyPath: String): ExtendedPublicKey = derivePublicKey(KeyPath.fromPath(keyPath))
 
-        public fun fingerprint(): Long = Pack.int32LE(ByteArrayInput(Crypto.hash160(publickeybytes).take(4).reversed().toByteArray())).toLong()
```
t-bast, line 177: We were indeed doing some very weird by-hand big-endian from little-endian, not sure why we did that in the first place...

src/commonTest/kotlin/fr/acinq/bitcoin/DeterministicWalletTestsCommon.kt
```diff
@@ -46,151 +70,91 @@ class DeterministicWalletTestsCommon {
             m_pub.encode(testnet = false),
             "xpub661MyMwAqRbcFtXgS5sYJABqqG9YLmC4Q1Rdap9gSE8NqtwybGhePY2gZ29ESFjqJoCu1Rupje8YtGqsefD265TMg7usUDFdp6W1EGMcet8"
         )
-        assertEquals(m.fingerprint(), 876747070L)
+        assertEquals(0x3442193E, m.fingerprint())
```
t-bast, line 73: I'm not seeing any *new* test case that failed with the previous code? We definitely need one, please let me know which one it is if I missed it, or add one before we merge?

t-bast (commented):
Good clean-up!

src/commonTest/kotlin/fr/acinq/bitcoin/DeterministicWalletTestsCommon.kt
sstone: tests in https://github.com/ACINQ/bitcoin-kmp/blob/ea57c7318013abfb1c067b1a2da8330258fb1415/src/commonTest/kotlin/fr/acinq/bitcoin/DeriveWalletKeysTestsCommon.kt now include explicit fingerprint verification and will fail with the previous `fingerprint()` method

t-bast: I believe you mean in `DeterministicWalletTestsCommon.kt`? I've been able to verify on `master` than indeed, starting with `m0h_1` and its children, the test fails because of a fingerprint mismatch.

t-bast (approved):

PR #182, opened 2026-05-19
Build cleanup (no functional changes)

sstone:
Cleanup of build and publishing scripts, no functional changes.

PR #183, opened 2026-05-19
Performance improvements (no functional changes) 

sstone:
2 simple performance improvements:
- reduce the number of temporary byte buffers (created whenever we call `.toByteArray()` or during byte vector operations)
- compute intermediary hashes used in transaction hashes once when possible

Based on  https://github.com/ACINQ/bitcoin-kmp/pull/182, will be rebased when it's merged.


src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -16,6 +16,7 @@
 
 package fr.acinq.bitcoin
 
+import fr.acinq.bitcoin.crypto.Digest
```
t-bast, line 19: nit: unused

src/commonMain/kotlin/fr/acinq/bitcoin/XonlyPublicKey.kt
```diff
@@ -22,14 +22,19 @@ import kotlin.jvm.JvmField
  * we only store the x coordinate of the pubkey, the y coordinate is always even
  */
 public data class XonlyPublicKey(@JvmField val value: ByteVector32) {
-    public constructor(pub: PublicKey) : this(pub.value.drop(1).toByteArray().byteVector32())
+    public constructor(pub: PublicKey) : this(ByteVector32(pub.value.bytes, 1))
```
t-bast, line 25: Claude caught a dangerous bug here, this should be:


```suggestion
    public constructor(pub: PublicKey) : this(ByteVector32(pub.value.bytes, pub.value.offset + 1))
```

More details:

```
This hard-codes offset 1 and ignores pub.value.offset. If a PublicKey is ever constructed from a sliced ByteVector (offset ≠ 0) — for example via
PublicKey(it.key.drop(1)), which PSBT parsing does at psbt/Psbt.kt:921 and :948 — then .xOnly() silently returns the wrong 32 bytes (parity prefix + 31
bytes of the x-coordinate). The ByteVector32 length check still passes, so it fails silently. Existing tests don't catch it because they all build
PublicKeys via privateKey.publicKey() or PublicKey.fromHex(...), where value.offset == 0.
```

src/commonMain/kotlin/fr/acinq/bitcoin/PrivateKey.kt
```diff
@@ -30,8 +30,8 @@ import kotlin.jvm.JvmStatic
 public data class PrivateKey(@JvmField val value: ByteVector32) {
     public constructor(data: ByteArray) : this(
         when {
-            data.size == 32 -> ByteVector32(data.copyOf())
-            data.size == 33 && data.last() == 1.toByte() -> ByteVector32(data.copyOf(32))
+            data.size == 32 -> ByteVector32(data)
+            data.size == 33 && data.last() == 1.toByte() -> ByteVector32(data)
```
t-bast, line 34: I think we should keep copying here (defensive copy is a best practice for key material), to ensure that memory can be freed. It's especially weird in the 33-bytes case to keep a reference to a 33-byte `ByteArray` inside a `ByteVector32` object...

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -86,7 +86,7 @@ public object Script {
     public fun parse(blob: ByteArray): List<ScriptElt> = parse(ByteArrayInput(blob))
 
     @JvmStatic
-    public fun parse(blob: ByteVector): List<ScriptElt> = parse(blob.toByteArray())
+    public fun parse(blob: ByteVector): List<ScriptElt> = parse(ByteArrayInput(blob.bytes, blob.offset))
```
t-bast, line 89: This one is problematic as well if we ever pass a sliced `ByteVector` as input (which we don't internally, but could happen in a future refactoring or could be done by an external caller). Full details (courtesy of Claude):

```
  Commit 7cd9676 rewrote the ByteVector overload of Script.parse from:                                                                                   
                                                                                                                                                         
  public fun parse(blob: ByteVector): List<ScriptElt> = parse(blob.toByteArray())                                                                        
                                                                                                                                                         
  to:
                                                                                                                                                         
  public fun parse(blob: ByteVector): List<ScriptElt> = parse(ByteArrayInput(blob.bytes, blob.offset))
                                                                                                                                                         
  The motivation is obvious — avoid the toByteArray() copy. But the substitution isn't actually equivalent.                                              
                                                                                                                                                         
  Why it's broken in principle                                                                                                                           
                                                        
  ByteArrayInput (io/ByteArrayInput.kt) now looks like:                                                                                                  
                                                        
  public class ByteArrayInput(private val array: ByteArray, offset: Int) : Input {                                                                       
      public constructor(array: ByteArray) : this(array, 0)
      private var position: Int = offset                                                                                                                 
      override val availableBytes: Int get() = array.size - position
      ...                                                                                                                                                
  }                                                     

  It tracks a single position and reports array.size - position as availableBytes. There is no notion of an upper bound — it reads until the end of the  
  underlying array.
                                                                                                                                                         
  And Script.parse(Input) uses exactly that field to decide when to stop:

  override fun hasNext(): Boolean = input.availableBytes > 0
                                                                                                                                                         
  So when Script.parse(blob) is given a ByteVector whose underlying bytes array is larger than the script (i.e. blob.offset + blob.size() <              
  blob.bytes.size), the iterator keeps consuming bytes past the end of the script's logical slice. Those extra bytes belong to whatever happens to come  
  after the slice in the buffer — random data — which the script parser will happily interpret as more opcodes / push-data.                              
                                                        
  A worked example:
  - blob.bytes = a 100-byte buffer.
  - blob.offset = 10, blob.size() = 5 — the script is 5 bytes long, sitting at indices 10..14.                                                           
  - Old code: blob.toByteArray() → fresh 5-byte array, parser reads exactly 5 bytes and stops. ✓
  - New code: ByteArrayInput(blob.bytes, 10) → availableBytes = 100 - 10 = 90. The parser will keep going past index 14 and parse bytes 15..99 as        
  additional script elements. ✗                                                                                                                          
                                                                                                                                                         
  Worse, if one of those extra bytes is a OP_PUSHDATA opcode with a large length prefix (e.g. 0x4e = OP_PUSHDATA4), the parser will try to read a        
  multi-gigabyte push and either OOM or throw, depending on what's available.                                                                            
                                                        
  Why no test fails today                                                                                                                                
                                                        
  The codebase happens to only ever pass un-sliced ByteVectors to Script.parse. I traced every call site:                                                
                                                        
  ┌────────────────────────────────────────┬───────────────────────────────┐                                                                             
  │               Call site                │   Source of the ByteVector    │
  ├────────────────────────────────────────┼───────────────────────────────┤
  │ Bitcoin.kt:172                         │ TxOut.publicKeyScript         │
  ├────────────────────────────────────────┼───────────────────────────────┤
  │ Transaction.kt:474                     │ previousOutputScript argument │                                                                             
  ├────────────────────────────────────────┼───────────────────────────────┤
  │ psbt/Psbt.kt:383, 403                  │ TxOut.publicKeyScript         │                                                                             
  ├────────────────────────────────────────┼───────────────────────────────┤                                                                             
  │ psbt/Psbt.kt:934, 940, 959, 1059, 1065 │ DataEntry.value               │
  └────────────────────────────────────────┴───────────────────────────────┘                                                                             
                                                        
  In every case the ByteVector was built from a fresh ByteArray:                                                                                         
  - TxOut's publicKeyScript comes from BtcSerializer.script(input) which returns bytes(input, length) — a freshly allocated array of exactly the script
  length, then wrapped in ByteVector(...) with offset 0.                                                                                                 
  - DataEntry's value comes from input.readNBytes(valueLength) (Psbt.kt:1191) — same story, fresh array.
                                                                                                                                                         
  So blob.bytes.size == blob.size() and blob.offset == 0 for every actual call, and the bug is invisible at runtime.                                     
                                                                                                                                                         
  Why it still matters                                                                                                                                   
                                                                                                                                                         
  1. Script.parse(blob: ByteVector) is a public API (@JvmStatic). Any external user passing a sliced ByteVector — including a perfectly natural          
  someBytes.drop(n) or take(n) — gets garbage results. The function's contract is "parse this script"; the implementation now parses "this script plus
  whatever happens to be after it in memory".                                                                                                            
  2. The codebase is one refactor away from triggering it internally. If someone someday reads multiple scripts out of a larger buffer using drop/slice
  instead of copying, the parser will silently misbehave. There's nothing in the type system or the function signature warning them off.                 
  3. It's a silent corruption bug, not a crash. When it does trigger, parsing succeeds and returns a List<ScriptElt> containing both the real script and
  trailing junk. Downstream code (signature checking, address derivation, etc.) then operates on wrong data.                                             
                                                        
  How to fix                                                                                                                                             
                                                        
  Three reasonable options, roughly in order of preference:                                                                                              
  
  1. Add a length-aware ByteArrayInput constructor and use it here. Add ByteArrayInput(array: ByteArray, offset: Int, length: Int) that bounds           
  availableBytes to (offset + length) - position. Then: 
  public fun parse(blob: ByteVector): List<ScriptElt> =                                                                                                  
      parse(ByteArrayInput(blob.bytes, blob.offset, blob.size()))                                                                                        
  1. This is the cleanest — it preserves the zero-copy goal of the commit and fixes the bug.
  2. Bound the script parser by the input's known length rather than availableBytes. More invasive — touches the script iterator — and doesn't           
  generalize.                                                                                                                                            
  3. Revert this one line to parse(blob.toByteArray()). Loses the optimization but is a one-line change and undeniably correct. Script blobs are         
  typically small (< 10 KB), so the copy cost is negligible compared to what Script.parse itself does.                                                   
                                                        
  I'd recommend option 1: it generalizes the Input abstraction in a way that matches the rest of the codebase's "byte buffer + offset + length"          
  convention (ByteVector, Crypto.sha256(bytes, offset, len), etc.), and it lets you keep the perf win.
```

t-bast (commented):
This is honestly a dangerous set of code changes...I understand the motivation, but it needs some real auditing and hardening if we want to go down that road. A simple first pass with Claude highlighted issues already, so it's worth digging more. If we want to merge this.

sstone:
replaced by https://github.com/ACINQ/bitcoin-kmp/pull/184

PR #184, opened 2026-05-19
Performance improvements 

sstone:
Simpler alternative to https://github.com/ACINQ/bitcoin-kmp/pull/183, which keeps improvements on transaction hashing but discard potentially dangerous byte vector optimisations.


src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
```diff
@@ -309,10 +310,55 @@ public data class Transaction(
     @JvmField val lockTime: Long
 ) : BtcSerializable<Transaction> {
 
+    private val hashPrevOut: ByteArray by lazy {
```
t-bast, line 316: Can you add a comment to explain why we're doing this (hash only once because this is independent of the input and otherwise done multiple times for no good reason if I understood correctly)?

t-bast (commented):
Looks good, this is safer than the alternative PR.

src/commonMain/kotlin/fr/acinq/bitcoin/Transaction.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/184/commits/777a5058c09e991ebd06b9f858b2f814405728b5

t-bast (approved):

PR #189, opened 2026-08-03
Performance improvements and bug fixes

sstone:
Commits are small and independent , see commit messages for more details:

- fix for an error in our native implementation of pbkdf2 (i.e. used only on linux and macos, not on JVM/Android/iOS) and only for key lengths that are not used in any of our products or any bitcoin-related process that I know of (i.e. seed generation, for example,  is correct on all platforms)
- add small performance improvements to our script engine
- improves the "already consumed" guard on `SecretNonce` and makes it atomic
- improves and simplifies our CI scripts
- adds all the official BIP39 wordlists to our unit tests
- rejects non-ascii BIP39 passphrases. This does not impact phoenix or phoenixd as they only use empty passphrases)  

This PR does not introduce backwards-compatibility issues for any of our products:

The pbkdf2 issue was specific to macos/linux (i.e Phoenix  iOS/Android was not affected), and only for key lengths that we don't use (and that no bitcoin algorithm would use AFAIK)

The BIP39 seed generation issue that we fix is specific to the JVM and non-ascii wordlists. Phoenix Android only uses the English BIP39 word list and is not affected.

Rejecting non-ascii BIP39 passphrases has not impact on Phoenix (iOS/Android) which only supports empty passphrases.


src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -201,6 +201,7 @@ public data class BlockHeader(
 public object MerkleTree {
     @JvmStatic
     public tailrec fun computeRoot(tree: List<ByteVector32>): ByteVector32 {
+        require(tree.isNotEmpty())
```
t-bast: nit: maybe worth adding a message?

```
require(tree.isNotEmpty()) { "cannot compute root of empty merkle tree" }
```

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
```diff
@@ -261,6 +262,7 @@ public data class Block(@JvmField val header: BlockHeader, @JvmField val tx: Lis
         override fun validate(message: Block) {
             BlockHeader.validate(message.header)
             require(message.header.hashMerkleRoot == MerkleTree.computeRoot(message.tx.map { it.hash.value })) { "invalid block:  merkle root mismatch" }
+            require(message.tx.isNotEmpty()) { "invalid block: no transactions" }
```
t-bast, line 264: nit: it would make a bit more sense to make this requirement *before* trying to compute the merkle root, the error message would be more helpful that way

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
```diff
@@ -884,6 +884,32 @@ public object Script {
             // OP_ENDIF
             // OP_CHECKSIG // conditions = []
             val conditions = mutableListOf<Boolean>()
+            // Number of `false` entries in `conditions`, maintained incrementally so that testing whether we're inside
```
t-bast, line 888: Do we have test coverage for the code modified here? I find it always hard to review changes to the script interpreter, it's tough to ensure that we're not diverging from bitcoind.

src/nativeMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.native.kt
```diff
@@ -9,17 +9,45 @@ public object Pbkdf2Native {
         fun process(input: ByteArray): ByteArray
     }
 
-    private class Hmac512(public val password: ByteArray) : Prf {
-        val digest: Digest = Digest.sha512()
+    private class Hmac512(password: ByteArray) : Prf {
```
t-bast, line 12: I've probably already said it the first time we introduced this, but it's never a good idea to roll our own crypto here. I have no idea whether the updated code is really 100% correct and safe. Isn't there any way to avoid this entirely?

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/189/changes/e552f069af1fe13c56418d893e3800137a5cd66d

src/commonMain/kotlin/fr/acinq/bitcoin/Block.kt
sstone: done in https://github.com/ACINQ/bitcoin-kmp/pull/189/changes/e552f069af1fe13c56418d893e3800137a5cd66d

src/commonMain/kotlin/fr/acinq/bitcoin/Script.kt
sstone: Yes it's covered by the reference script tests we import from bitcoin-core.

src/nativeMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.native.kt
sstone: We don't have much of a choice as there is no widely used kotlin code for this. The only option I can think of would be to reuse C code from a widely used library and write a JNI wrapper but it would be overkill. 
The code here is short and fairly simple, the error we had was not detected because it happens with parameters that we never use and that no bitcoin standard would use afaik so our bitcoin-specific tests did not catch it.

t-bast: Ok, I'll spend more time reviewing this in details then to get more confidence that this is correct.

src/iosMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.ios.kt
```diff
@@ -12,6 +12,8 @@ public actual object Pbkdf2 {
 
     @OptIn(ExperimentalUnsignedTypes::class, ExperimentalForeignApi::class)
     public actual fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray {
+        require(count >= 1) { "iteration count must be greater than 0" }
+        require(dkLen >= 1) { "derived key length must be greater than 0" }
         memScoped {
             val result = ByteArray(dkLen)
             KCCKeyDerivationPBKDF(
```
t-bast: You're not checking the return value of this function. According to Opus 5 this is an issue, because on `kCCParamError` the function silently returns an all-zero key. We should add a require:

```
val status = KCCKeyDerivationPBKDF(...)
require(status == 0) { "CCKeyDerivationPBKDF failed with status $status" }
```

src/iosMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.ios.kt
t-bast, line 26: The behavior of those is unclear if `password` or `salt` is empty. Shouldn't we `require` that both of those aren't empty?

src/iosMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.ios.kt
```diff
@@ -12,9 +12,14 @@ public actual object Pbkdf2 {
 
     @OptIn(ExperimentalUnsignedTypes::class, ExperimentalForeignApi::class)
     public actual fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray {
+        require(count >= 1) { "iteration count must be greater than 0" }
+        require(dkLen >= 1) { "derived key length must be greater than 0" }
+        require(password.isNotEmpty()) { "password must not be empty" }
+        require(salt.isNotEmpty()) { "password must not be empty" }
```
t-bast:
```suggestion
        require(salt.isNotEmpty()) { "salt must not be empty" }
```

src/jvmMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.jvm.kt
```diff
@@ -7,8 +7,12 @@ import javax.crypto.spec.PBEKeySpec
 public actual object Pbkdf2 {
 
     @JvmStatic
-    public actual fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray =
-        SecretKeyFactory.getInstance("PBKDF2WithHmacSHA512")
+    public actual fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray {
+        require(password.isNotEmpty()) { "password must not be empty" }
+        require(salt.isNotEmpty()) { "password must not be empty" }
```
t-bast:
```suggestion
        require(salt.isNotEmpty()) { "salt must not be empty" }
```

src/nativeMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.native.kt
```diff
@@ -39,13 +67,19 @@ public object Pbkdf2Native {
...
+            f(i).copyInto(t, destinationOffset = (i - 1) * hLen, startIndex = 0, endIndex = if (i == l) r else hLen)
         }
         return t
     }
 
-    public fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray =
-        generate(salt, count, dkLen, Hmac512(password))
+    public fun withHmacSha512(password: ByteArray, salt: ByteArray, count: Int, dkLen: Int): ByteArray {
+        require(password.isNotEmpty()) { "password must not be empty" }
+        require(salt.isNotEmpty()) { "password must not be empty" }
```
t-bast:
```suggestion
        require(salt.isNotEmpty()) { "salt must not be empty" }
```

src/jvmMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.jvm.kt
sstone: oops ... fixed in https://github.com/ACINQ/bitcoin-kmp/pull/189/commits/a7569929c9c4a0d5e5b02a0ef59da11bc63fa8eb

src/nativeMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.native.kt
sstone: fixed in https://github.com/ACINQ/bitcoin-kmp/pull/189/commits/a7569929c9c4a0d5e5b02a0ef59da11bc63fa8eb

src/iosMain/kotlin/fr/acinq/bitcoin/crypto/pbkdf2.ios.kt
sstone: fixed in https://github.com/ACINQ/bitcoin-kmp/pull/189/commits/a7569929c9c4a0d5e5b02a0ef59da11bc63fa8eb

t-bast (commented):
LGTM! Since we're touching dangerous parts of the core crypto with the seed stuff, we should do an e2e seed restore test on android/iOS/linux (for phoenixd) to validate this before actually making a release of `bitcoin-kmp` that uses this.

t-bast (commented):
Can you summarize where we're at regarding the encoding issue for PBKDF2?

If I understand correctly, the `password` that we're feeding into PBKDF2 is the seed words and the `salt` that we're providing is always just "mnemonic" (we don't support adding a passphrase in phoenix, do we?).

The JVM code (used for Phoenix android) was previously invalid if fed non-ascii characters, and is updated in this PR to correctly support arbitrary UTF8 characters: this is backwards-compatible because we currently only support english seed words on Phoenix android? Which means that after this change, a chinese seed created on iOS would correctly match when imported into android (and previously wouldn't match)?

The iOS and native code was previously already correctly supporting arbitrary UTF-8 characters and is unchanged, is that correct?

So after this change, Phoenix android and iOS will correctly generate the same seed even when using non-english wordlists? And we're not introducing any backwards-compatibility issue by changing the derived seed of existing users?

sstone:
> Can you summarize where we're at regarding the encoding issue for PBKDF2?
> 
> If I understand correctly, the `password` that we're feeding into PBKDF2 is the seed words and the `salt` that we're providing is always just "mnemonic" (we don't support adding a passphrase in phoenix, do we?).
> 
> The JVM code (used for Phoenix android) was previously invalid if fed non-ascii characters, and is updated in this PR to correctly support arbitrary UTF8 characters: this is backwards-compatible because we currently only support english seed words on Phoenix android? Which means that after this change, a chinese seed created on iOS would correctly match when imported into android (and previously wouldn't match)?
> 
> The iOS and native code was previously already correctly supporting arbitrary UTF-8 characters and is unchanged, is that correct?
> 
> So after this change, Phoenix android and iOS will correctly generate the same seed even when using non-english wordlists? And we're not introducing any backwards-compatibility issue by changing the derived seed of existing users?

Yes sorry If I was unclear, I've updated the PR description.

t-bast (approved):
Great thanks for the details!

dpad85:
> we don't support adding a passphrase in phoenix, do we?

Correct. Passphrase is always empty. We supported passphrases in Eclair Mobile and that caused issues as many users were not aware that passphrase needed to be backed up as well.

> because we currently only support english seed words on Phoenix android?

Correct.

> a chinese seed created on iOS would correctly match when imported into android?

The iOS app only supports French, Spanish and Czech words list, in addition to English. Besides,"special" characters are stripped from the words list for those languages (see https://github.com/ACINQ/phoenix/blob/3749b0817d60b0a17a97d21338e2d9038efb2d9e/phoenix-shared/src/commonMain/kotlin/fr.acinq.phoenix/utils/MnemonicLanguage.kt#L5).

Also, the Phoenix Android app will not let you import a non-English seed.

PR #191, opened 2026-08-25
Small bug fixes related to script flags and script execution

sstone:
This PR fixes minor bugs in our script engines (see each commit description for more details), which affect non-standard p2sh transactions or specific combination of script execution flags.
None of these bugs apply to scripts used in lightning, or standard (wallet-generated etc... ) scripts but they may impact some users of this library.
I've built and tested `eclair` and `lightning-kmp` with these changes without issues. 
SNAPSHOTs are published to https://central.sonatype.com/repository/maven-snapshots/


t-bast (approved):
I have honestly no idea whether https://github.com/ACINQ/bitcoin-kmp/pull/191/changes/a907d891519b77724f5e78838a65813a092f7cd1 and https://github.com/ACINQ/bitcoin-kmp/pull/191/changes/f304187de5e3dfebb04b45c13a0bc85de3f23748 are doing what we want and not introducing subtle issues. It looks ok (maybe) but it's really hard to tell so I'm just trusting the test vectors here, my review isn't providing much security.

sstone:
> I have honestly no idea whether [a907d89](https://github.com/ACINQ/bitcoin-kmp/commit/a907d891519b77724f5e78838a65813a092f7cd1) and [f304187](https://github.com/ACINQ/bitcoin-kmp/commit/f304187de5e3dfebb04b45c13a0bc85de3f23748) are doing what we want and not introducing subtle issues. It looks ok (maybe) but it's really hard to tell so I'm just trusting the test vectors here, my review isn't providing much security.

[f304187](https://github.com/ACINQ/bitcoin-kmp/commit/f304187de5e3dfebb04b45c13a0bc85de3f23748)  fixes an issue for a combination of script flags that we don't use (and most bitcoin apps would not use).

[a907d89](https://github.com/ACINQ/bitcoin-kmp/commit/a907d891519b77724f5e78838a65813a092f7cd1) fixes an issue that is not covered by bitcoin core's reference test vectors which is why I added some with the same format. our version of `verifyScripts()` is updated and is much closer to bitcoin core's which will help future changes and reviews.

t-bast:
What I meant is that I understand the purpose, but the code is honestly extremely hard to review for correctness - especially since what we need to review is that the behavior matches bitcoin core.

sstone:
> What I meant is that I understand the purpose, but the code is honestly extremely hard to review for correctness - especially since what we need to review is that the behavior matches bitcoin core.

Yes, these changes cannot be reviewed on their own, you need to compare against bitcoin core code side-by-side.

PR #192, opened 2026-09-03
Don't append a SIGHASH_DEFAULT byte to taproot key path signatures

gorunjinian:
`Psbt.signWitness` appends the sighash byte whenever `input.sighashType` is non-null, including when it is `SIGHASH_DEFAULT` (`0x00`).

BIP-341 requires that a 65-byte signature's final byte not be `0x00`, so 64-byte signatures can't be malleated into 65-byte ones. This library already enforces that in `Script.sigHashType`, which raises `error("invalid sig hashtype")` on exactly what the signer produces here — so the signer emits a signature its own verifier (and consensus) rejects.

It's reachable through the public API: `updateWitnessInput(…, sighashType = SIGHASH_DEFAULT)` stores the value without validation, and parsing yields a non-null `0` whenever `PSBT_IN_SIGHASH_TYPE` is explicitly present.

The fix reads the sighash type once and appends the byte only for non-default types. The added test fails on master with:

```
java.lang.AssertionError: expected:<64> but was:<65>
```


sstone:
Needs rebase.

sstone (approved):

PR #193, opened 2026-09-03
Parse the taproot BIP32 derivation fingerprint as an unsigned value

gorunjinian:
`TaprootBip32DerivationPath.read` parses the master key fingerprint with `Pack.int32BE(input).toLong()`, which sign-extends it. The three other fingerprint parse sites in the same file — `PSBT_GLOBAL_XPUB`, `PSBT_IN_BIP32_DERIVATION` and `PSBT_OUT_BIP32_DERIVATION` — all use `.toUInt().toLong()`; only the taproot path differs.

Fingerprints are unsigned 32-bit values, so roughly half have a first byte `>= 0x80`. For those, `PSBT_IN_TAP_BIP32_DERIVATION` decodes to a negative `Long` that never equals the fingerprint a wallet computes for itself, so it doesn't recognise its own taproot inputs. In practice a BIP-86 wallet can't sign its own PSBTs while the same key material signs native segwit PSBTs fine.

The existing taproot-derivation fixtures use fingerprints `0x772b2da7` and `0`, both with the high bit clear, which is why the current suite doesn't catch it. The added test covers the direct `write()`/`read()` round trip and a full `Psbt.write`/`Psbt.read` round trip with fingerprint `0x8dfc9b34`; it fails on master with:

```
java.lang.AssertionError: expected:<2382142260> but was:<-1912825036>
```

### DIFF
--- a/src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
+++ b/src/commonMain/kotlin/fr/acinq/bitcoin/psbt/Psbt.kt
@@ -1227,7 +1227,10 @@ public data class TaprootBip32DerivationPath(@JvmField val leaves: List<ByteVect
             val input = ByteArrayInput(bin)
             val numLeaves = BtcSerializer.varint(input).toInt()
             val leaves = (0 until numLeaves).map { BtcSerializer.bytes(input, 32).byteVector32() }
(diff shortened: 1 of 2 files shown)


sstone (approved):
Thanks!

